PR #22, opened 2016-03-29
Add --wait to explicitly wait for the roscore

Rayman:
This makes it so you HAVE to have a separate roscore. Is this what we want?

@jlunenburg, @svddries 
Fix tue-robotics/tue_robocup#81


Timple:
Solved and answered by https://github.com/tue-robotics/amigo_bringup/pull/9

PR #25, opened 2016-09-16
[To Test] Convert all navigation targets to yaml

Rayman:


MatthijsBurgh:
Looks like everything is the same except the hydro-level branche of the arm_navigation_msgs target is missing. If that is correct, I say merge!

PR #27, opened 2016-11-29
use catkin build instead of catkin_make

Timple:
I would propose to use -by default- catkin build instead of catkin_make

tue-make still supports the (current) use of catkin_make

Rationale can be found here: https://catkin-tools.readthedocs.io/en/latest/history.html

Anyone to argue for or against this?


Timple:
Seems to work and is actually required for amigo1, so merging

PR #38, opened 2017-02-01
Fixing robocup bash functions for new folder stucture

MatthijsBurgh:
Work in progress

### DIFF
--- a/setup/tue.bash_functions
+++ b/setup/tue.bash_functions
@@ -666,11 +666,16 @@ For example:
 
     cd ~/.tue
     git remote set-url $remote ${server}tue-robotics/tue-env
-
-    local fs=`ls $TUE_ENV_DIR/repos/https:/github.com/tue-robotics`
+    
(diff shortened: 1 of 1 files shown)


setup/tue.bash_functions
```diff
@@ -719,7 +729,7 @@ function tue-robocup-update
     cd ~/.tue
     git pull --ff-only
 
-    tue-robocup-install-package picaso_4d_systems
+    tue-robocup-install-package picaso_4d_systems #WHY????
```
LoyVanBeek: Indeed, why? 
This line is added in https://github.com/tue-robotics/tue-env/commit/2fa40d7a3b818fac127963a5ce801462495495aa by @svddries and I think it was for a touch screen we had back then. Relatively sure this is OK for removal again.

reinzor: Yes, can be removed .. We are not using this anymore and there is no use for that line being there

MatthijsBurgh:
@LoyVanBeek or @reinzor, Please a final review

LoyVanBeek:
I don't see anything to disagree with :-)

PR #44, opened 2017-02-13
github-cache-credentials function

MatthijsBurgh:
testing is difficult, but you can use echo to see if the git line is correct.


installer/targets/robot-user/setup
```diff
@@ -124,4 +159,3 @@ if [ ! $ROBOT_REAL ] ; then
 		export ROS_IP=`hostname -I | sed 's/.*\(192\.168\.2\.[0-9]\+\).*/\1/'`
 	fi
 fi
-
```
reinzor: We always want a newline at the end of a file to make the linters happy

PR #47, opened 2017-02-15
Fix/target soem beckhoff drivers

MatthijsBurgh:


rokusottervanger:
Looks good to me, but Rein worked on refactoring targets, so he should probably review this.

PR #49, opened 2017-02-15
refactor(): remove all astart and amiddle functions

reinzor:
No specific challenge functions and/or different set envs


LoyVanBeek:
What are the replacements, if needed? Better if they aren't needed, but I suppose we had them for a reason.

rokusottervanger:
This was a quick RoboCup hack to start the right world model configuration for every challenge. In my opinion we should find a more scalable solution. Maybe the challenge should configure the world model or something like that.

reinzor:
I would like to work to a solution where we start only the challenge launch files and that includes also the specific deps. So the start is running on amigo1 by default and a second launch file for the challenge and deps

PR #64, opened 2017-03-11
sshamigoX not using -X to keep sound at remote

reinzor:


LoyVanBeek:
This means no more X11 forwarding, so no graphical editors to edit stuff on Amigo. 
VIM FTW!

PR #71, opened 2017-04-08
clean up of sergio1,2 target, similar to amigo1,2 target

MaxBaeten:


MaxBaeten:
And can we also remove the amigo and sergio targets? since they now have specific targets for each pc?

MaxBaeten:
install.bash is empty for both the amigo and sergio targets, can these files be removed?

reinzor:
Yes 

-Rein

PR #74, opened 2017-04-17
removing amigo3 en sergio3 from multiple functions

MatthijsBurgh:


MaxBaeten:
shall we also remove the -X from the ado and sdo functions? These are never used to open gui's anyway.

MatthijsBurgh:
I think that is a good idea. Do we keep the -X in the terminator configs?

MaxBaeten:
No you can also remove them there.

MatthijsBurgh:
I think it ready to be reviewed/merged

Timple:
Is there a benefit of removing the -X? In the past it was sometimes used
for geany to modify config files on Amigo

Op 17 apr. 2017 6:56 p.m. schreef "Max Baeten" <notifications@github.com>:

MatthijsBurgh:
This can still be done manually

PR #86, opened 2017-05-26
Add my work email so I can git su for that as well

LoyVanBeek:


MatthijsBurgh:
How does git su handle two identical names?

LoyVanBeek:
Dunno, this kinda works. 'loy' resolves to my gmail (I guess because its listed first) and 'nobleo' resolves to my work email.

LoyVanBeek:
Also, if there is a clash, I should be the only one ever suffering it.

PR #93, opened 2017-06-03
Removing robot_smach_states install.bash target

MatthijsBurgh:
Already in https://github.com/tue-robotics/tue-env/blob/master/installer/scripts/tue-install-impl#L661


MatthijsBurgh:
It also in the tensorflow target, but it needs to stay there,  because bash files are sourced before pip check is run in the installer.

MatthijsBurgh:
Combined pull request with, tue-robotics/tue_robocup#309

MatthijsBurgh:
Please don't merge, I will do a final test before merging.

MatthijsBurgh:
No problems during tue-get update, clean install difficult to test. But no typo's etc.

LoyVanBeek:
Can you check in a docker? Based on tue-env?

MatthijsBurgh:
I am not familiar with docker. Do you have a tutorial of sample?

LoyVanBeek:
I see installation of orocos-kdl and python-orocos-kdl is also removed. Are those installed via another way?

MatthijsBurgh:
These are add to package.xml. See pull request in tue_robocup

PR #97, opened 2017-06-08
Add support for pull requests instead of only compiling branches

Timple:
Works, see https://github.com/tue-robotics/amigo_bringup/pull/47

### DIFF
--- a/ci/install-package.sh
+++ b/ci/install-package.sh
@@ -1,24 +1,25 @@
 #!/bin/bash
 set -o errexit #Stop on errors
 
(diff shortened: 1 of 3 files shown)


ci/install-package.sh
```diff
@@ -16,21 +16,26 @@ case $i in
     COMMIT="${i#*=}"
     shift # past argument=value
     ;;
+    -r=*|--pullrequest=*)
```
LoyVanBeek: Some docs would be nice, I have no idea what is going on here.

PR #100, opened 2017-06-09
Include running tests on compiling

Timple:


ci/build-package.sh
```diff
@@ -20,3 +20,7 @@ echo -e "\e[35m\e[1m PACKAGE     = ${PACKAGE} \e[0m"
 # Compile the package
 echo -e "\e[35m\e[1m Compile the package \e[0m"
 docker exec tue-env bash -c "export CI='true'; source /home/amigo/.bashrc; cd ~/ros/kinetic/system/src/$PACKAGE && catkin build --this --no-status"
+
+# Run unit tests
+echo -e "\e[35m\e[1m Run unit test on this package (catkin run_test --this --no-deps \e[0m"
+docker exec tue-env bash -c "export CI='true'; source /home/amigo/.bashrc; cd ~/ros/kinetic/system/src/$PACKAGE && catkin run_tests --this --no-status --no-deps"
```
LoyVanBeek: Perhaps put this in a separate script (ci/test-package.sh) so that in the travis output, we can see that build went OK but the tests failed, which is a useful distinction.

Timple: I thought about it. But it is still mostly compilation stuff.

I want to make a seperate script for launching stuff and for simulating stuff.

Too many distinctions will also clutter the travis.yml files everywhere which is exactly what these scripts try to prevent. That the cmake tests fail instead of the build should be obvious from the highlighted comments.

PR #104, opened 2017-06-13
Stop on errors in install and bootstrap scripts. No use in going on

Timple:
This gives the user a change to fix it before a lot of potential useless commands are executed


Rayman:
This PR is broken on my machine. Revert PR here: #116

LoyVanBeek:
Stopping on error is OK I suppose, as long as the errors are correctly reported so they can be fixed. What errors are reported, if any?

PR #114, opened 2017-06-27
fix(tue-checkout): Also checkout .tue dir

reinzor:


setup/tue.bash_functions
```diff
@@ -568,6 +568,7 @@ function tue-checkout
     fi
 
     local branch=$1
+	cd $TUE_DIR; git checkout $branch > /dev/null; cd - > /dev/null
```
MatthijsBurgh: Please use same implementation as pkgs

reinzor:
Why?

-Rein

MatthijsBurgh:
At least the visuals would be useful

MatthijsBurgh:
I want to extend this with the option to disable the checkout of the .tue dir.

MatthijsBurgh:
@Timple could you create an option to not checkout the .tue dir? Like --only-pkgs.

MatthijsBurgh:
@Timple please review

setup/tue.bash_functions
```diff
@@ -561,18 +561,43 @@ function tue-checkout
     then
         echo """Switches all packages to the given branch, if such a branch exists in that package. Usage:
 
-    tue-branch BRANCH-NAME
+    tue-checkout BRANCH-NAME [option]
+
+    options:
+    --only-pks: tue-env is not checkout to the specified branch
```
LoyVanBeek: *check**ed** out

PR #116, opened 2017-07-04
Revert "Stop on errors in install and bootstrap scripts"

Rayman:
It breaks `tue-get update` on my machine

### DIFF
--- a/ci/build-docker-image.sh
+++ b/ci/build-docker-image.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -o errexit # Stop on errors
+set -o errexit #Stop on errors
 
 # create tag based on branch name
 export IMAGE_NAME=tuerobotics/tue-env:`echo "$TRAVIS_BRANCH" | tr '[:upper:]' '[:lower:]' | sed -e 's:/:_:g'`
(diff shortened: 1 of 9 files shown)


Timple:
Hoho, wait. Can we get a l ittle bit more info? Instead of just 'it breaks'?

There was a reason for implementing it. It prevents travis (and your local PC) from spitting out a whole bunch of unrelated errors if something goes wrong. What is not the proposed way of dealing with this?

Also regarding the reviewer, is this all info required for getting a pull request accepted? :P

MatthijsBurgh:
The stops for travis can be build in again. The normal installer should keep going in some way

LoyVanBeek:
No, I disagree: if your install(er) breaks for whatever reason, the error needs to be fixed, not ignored. That is asking for trouble.

Timple:
I cannot really think of an example where a failed command leads to the desired outcome.

@Rayman: Can you perhaps post the output? Of the command that fails (but is ok to do so). Then I will fix that in the scripts.

Rayman:
The problem is the follwoing: `tue-get update` does not work anymore. There is NO output, it just stops somewhere in the middle. I didn't have time to look for a cause, but it is clearly a regression.

Timple:
Found the issue, when sourcing a script (source script.sh), the commands are executed as if they would have been inside this terminal. This also means the shell exits on an error.
Running a script (./script.sh) does not have this.

Todo: source only when required (environmental variables need to be kept). Otherwise run a script.
Only the scripts that run can then be errexit-ed

PR #122, opened 2017-07-08
convert pip and system target to lowercase

MatthijsBurgh:


LoyVanBeek:
For PIP packages, thi sis OK, but apt is case sensitive. 
E.g. PyQT5 can be installed by pyqt5 not

LoyVanBeek:
Wait a sec. Why is this needed at all, since pip is case insensitive? To compare targets with?

MatthijsBurgh:
Then we have to be sure all the targets are correct. I thought both apt and pip only have lowercase targets.

MatthijsBurgh:
Because both apt and pip are case-sensitive, this will break the installer. So closing this PR.

PR #123, opened 2017-07-11
Fix small tue-make-dev bug

Rayman:


MatthijsBurgh:
Should you not check both system and dev? Because you don't want the dev compiled by a different catkin tool

Rayman:
Each workspace is separate and could be build by a different tool I think.

Timple:
If it works I'm fine with it. 

If it does not work I would not invest any time in it. Seems like an odd situation.

PR #125, opened 2017-07-18
Fix/missing git + removing graveyard targets

MatthijsBurgh:


Timple:
Is there a reason for .git at the end?

MatthijsBurgh:
It is not needed, because it is working right now. just to be consistent

PR #126, opened 2017-07-18
add .git check to installer

MatthijsBurgh:


MatthijsBurgh:
This goes wrong when .git is partially added, but that shouldn't happen anyway.

MatthijsBurgh:
@Timple What do you think about this one?

Rayman:
Not every git repo ends with `.git`.

PR #128, opened 2017-07-24
Use debian target for audio

Timple:
Rokus' commit is accepted: https://github.com/ros-drivers/audio_common/pull/93

Debian build is from 0.3.1-0xenial-20170613-153511-0800

So seems merged but needs to be confirmed.

Lets wait until after RoboCup


installer/targets/ros-audio_capture/install.yaml
```diff
@@ -1,6 +1,5 @@
 - type: ros
   source:
     type: git
```
MatthijsBurgh: this line needs to removed

Timple: True!

MatthijsBurgh:
@Timple let's merge?

installer/targets/ros-audio_capture/install.bash
```diff
@@ -1,18 +1,18 @@
...
+#for dep in $deps
+#do
+#    if [ $dep != "Depends:" ] && [ $dep != "ros-$ROS_DISTRO-audio-capture" ] && [ $dep != "ros-$ROS_DISTRO-audio-common-msgs" ] && [ $dep != "ros-$ROS_DISTRO-roscpp" ]
+#    then
+#        #echo $dep
+#        targets="$targets $dep"
+#    fi
+#done
+##echo $targets
+#tue-install-system $targets
```
Rayman: Just remove the file if everything is commented.

MatthijsBurgh: I wanted it to keep in case the debian pkg is not correct, but that is fixed by a revert.

Rayman:
If you rebase the last 3 commits into 1 I'll accept this (I can show you tonight how to do this). Has this been tested?

PR #131, opened 2017-08-03
Nobleo backport

Timple:
Reviewing per commit might be easiest.


installer/targets/robot/install.bash
```diff
@@ -1,2 +1,16 @@
+#Do not require sudo
+if [ ! -f /etc/sudoers.d/sudo-no-password ]
```
MatthijsBurgh: I don't know if this is save

setup/tue.bash_functions
```diff
@@ -418,6 +418,13 @@ function tue-get
         return 1
     fi
 
+    #Create btrfs snapshot if possible:
```
MatthijsBurgh: I am not really familiar with btrfs, but now this is done with every tue-get function. I think that is unwanted. If I am wrong, please explain.

installer/scripts/tue-install-impl
```diff
@@ -303,10 +323,21 @@ function tue-install-add-text
...
+            echo -e "$text" | sudo tee --append $target_file
+        else
+            echo -e "$text" | tee --append $target_file
+        fi
     else
-        sed -e "/^$end_tag/r $source_file" -e "/^$begin_tag/,/^$end_tag/d" $target_file > $target_file.tmp
-        mv $target_file.tmp $target_file
+        if $root_required
+        then
+            sed -e "/^$end_tag/r $source_file" -e "/^$begin_tag/,/^$end_tag/d" $target_file | sudo tee $target_file.tmp
```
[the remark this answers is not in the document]
Timple: sudo sed -i might also be possible. But using tee also shows what is happening, is a bit nicer to create awareness running tue-get update. And it only shows once, so should not be annoying.

installer/targets/robot/install.bash
Timple: Well, you are right that modifying the sudoers file is quit tricky. But I think it is less tricky once automated than people modifying these files themselves.

Hasn't gone wrong yet at least here ;-)

Feel free to try on your PC.

setup/tue.bash_functions
Timple: I think it is exactly wanted to make a backup at every tue-get action. Otherwise which moments would you select would be wise? 

In btrfs backups are cheap by the way. They do not cost any time. And if nothing has changed, they do not cost any space. So running tue-get update 20 times in a row has no disadvantage ;-)

MatthijsBurgh: But now it is also done when you tue-get dep etc., right? I think that is unwanted. So only with install, update and remove. Do you agree?

Timple: OK, that makes sense :+1:

MatthijsBurgh:
@Timple Do you agree with me? After my explanation of the remove function? Plus I agree with Ramon that we don't want passwordless sudo.

Timple:
I don't really agree, the remove function sounds like it should do a lot more. If we rely on the poor implementation now, we cannot improve the function ;-)

MatthijsBurgh:
But removing a target does not directly imply removing anything. Because an other installed target could still depend on that target or dependencies of that target. So if you would have a good implementation. The remove function should be really extended. But I agree with you be enabling it also for remove for purpose in the future.

Timple:
Looks good to me, lets resolve the conflicts and merge

MatthijsBurgh:
I am not really sure about the showing the version in tue-get dep. Targets are shown multiple times, Sometimes really a lot of times. So accessing the package.xml each time, really slows down this function. So removing is the way to go in my opinion.

Timple:
Can be removed or made optional, I don't mind.

MatthijsBurgh:
Optional is ok for me.

Timple:
If you approve the last commit I'll merge

installer/targets/robot/install.bash
```diff
@@ -1,2 +1,9 @@
+### Make sure remote launching works ###
 tue-install-cp ros-bash-and-run.sh ~/.ros-bash-and-run.sh
 chmod +x ~/.ros-bash-and-run.sh
+
+### Allow user to access serial interfaces ###
```
Rayman: Which serial interfaces?

Timple: For example the hokuyo or (in the future) the dynamixels

MatthijsBurgh:
I also added the ppa functionality in an other PR. Which is more inline with out current implementation. So please your opinion on that one. Then we can choose between those two.

MatthijsBurgh:
@Rayman, could you review once more?

PR #138, opened 2017-08-30
tue-save-map update

MatthijsBurgh:


installer/targets/ros-map_server/install.yaml
```diff
@@ -4,3 +4,9 @@
     url: https://github.com/tue-robotics/navigation.git
     sub-dir: map_server
     version: kinetic-devel
+
+- type: target
```
jlunenburg: We have to check what this does. It might just happen that this installs a lot of stuff on the robot pcs that we actually don't want there...

MatthijsBurgh: So move to install.bash and surround with if robot!=real? The tue-save-map function can also be placed in an if robot!=real. So it is available on the robot. Good solution?

reinzor:
Not used

MatthijsBurgh:
That is why I have updated it, so it can be used

installer/targets/ros-map_server/install.yaml
MatthijsBurgh: @jlunenburg please read my previous comment

installer/targets/tue-save-map/setup
```diff
@@ -0,0 +1,102 @@
+#! /bin/bash
+
+function _parse_yaml_map_server {
```
LoyVanBeek: Why are functions and utilities like these locked up in tue-env? This is not an environment function but something for manipulating maps.

installer/targets/tue-save-map/setup
```diff
@@ -0,0 +1,102 @@
...
+   local prefix=$2
+   local s='[[:space:]]*' w='[a-zA-Z0-9_]*' fs=$(echo @|tr @ '\034')
+   sed -ne "s|^\($s\):|\1|" \
+        -e "s|^\($s\)\($w\)$s:$s[\"']\(.*\)[\"']$s\$|\1$fs\2$fs\3|p" \
+        -e "s|^\($s\)\($w\)$s:$s\(.*\)$s\$|\1$fs\2$fs\3|p"  $1 |
```
Rayman: What does this code do. Why? Why not Python?

MatthijsBurgh: Parsing yaml in bash, copied from stackoverflow

alberth: Given the flexibility of the yaml language, I doubt you can do that reliably, ie "working" thus relies on special case formatted lines (where "special case" may be the common form, so "special" isn't noticed often).

In general, I believe we should move away from bash towards Python. There are many more people that can read and hack Python, which makes this stuff easier to work on by more people. Also, Python 3 would be preferred. This code isn't used in ROS tools, and any Python 2 you write today will need to be reworked in 2 years.

Rayman: @alberth, as a team, we are following the ros guidelines: http://www.ros.org/reps/rep-0003.html. So for kinetic (the current platform) the targeted languages are as follows:
```
Targeted Languages:

C++11
Python 2.7
Python 3.5 not required, but testing against it is recommended
```

alberth: @Rayman  Fair enough

@MatthijsBurgh  Add the link as well then, for some future person that has to decode what the code does?
````
https://stackoverflow.com/questions/5014632/how-can-i-parse-a-yaml-file-from-a-linux-shell-script
````
(assuming this is the link you used)

MatthijsBurgh:
sorry, ignoring the rules for now.

Rayman:
You got 2 negative reviews and still merged???

PR #140, opened 2017-09-05
Fix tue-dev install due to two issues

Rayman:


Rayman:
I could not reproduce this issue on a fresh install :(

Timple:
The issue or the solution Matthijs suggested?

Because I'm fairly sure it worked ;)

Rayman:
The issue, I'll see next tuesday if the issue persists on that pc. Don't delete the branch please

PR #145, opened 2017-09-08
Refactor/remove some pip targets

Rayman:
Tested this with a clean install in virtualbox, amigo-start & amigo-free-mode start without errors. Also amigo-console works.


Timple:
Let's include those three in the CI then 😊

PR #147, opened 2017-09-11
Deleting targets

MatthijsBurgh:


MatthijsBurgh:
@Rayman could you review?

MatthijsBurgh:
No, not at the moment. So when we will use it, we move the repo back and we recreate the target.

PR #151, opened 2017-09-14
add ppa functionality incl. yaml

MatthijsBurgh:
Plus some targets converted.


Timple:
@MatthijsBurgh what did you think of my comment to enforce starting with ppa instead of leaving it out. This would be consistent with the .git enforcement.

MatthijsBurgh:
This was the easier option, because with ppa: the grep is not working. So the ppa: needs to be removed or the ':' should be converted to '.*'. I don't suggest the last option. But I agree with you that it is better to use ppa:.

Timple:
`grep -h "^deb.*${ppa#ppa:}" /etc/apt/sources.list.d/*`

Timple:
is PPA_ADDED actually cleared between runs?

MatthijsBurgh:
I don't think so. Oops

PR #158, opened 2017-09-22
Remove source files of dlib

MatthijsBurgh:


installer/targets/openface/dlib.bash
```diff
@@ -29,6 +29,9 @@ sudo sh -c 'echo "/usr/lib/openblas-base/" > /etc/ld.so.conf.d/libopenblas-base.
 #Finally, run ldconfig to update the cache.
 sudo ldconfig
 
+#Remove source files
+sudo rm -rf ~/src
```
Rayman: I put a lot of my own programs in `~/src` so this is obviously wrong.

MatthijsBurgh: Why not dowload the stuff in /tmp? The binaries are copied and the other files aren't used anymore, right?

LoyVanBeek:
Why delete the source files? I don't know how often we recompile this stuff. 

I'd rather have redistributeable binaries for the whole thing.

LoyVanBeek:
Ugh, this is a compile and forget package, if you want to recompile you have bigger issues I suppose.

PR #159, opened 2017-09-26
remove editors and chrome from tue-dev, moved to tue-dev-tools

MatthijsBurgh:


installer/targets/tue-dev/install.yaml
```diff
@@ -49,26 +49,29 @@
...
 
 # ED rviz plugin
 - type: target
   name: ros-ed_rviz_plugins
-  
+
 # HSR commands
 - type: target
   name: hsr-user
+
```
LoyVanBeek: Now you've just moved it. 

For the sake of compose-ability, make separate tue-dev and tue-tools targets, which are disjoint. For your matthijs target, you can include both but perhaps a ```tue-dev-full``` target also includes both tue-dev and tue-tools

PR #161, opened 2017-09-26
Update README.md

MatthijsBurgh:


alberth:
The comment "#tue-dev plus extra tools" implies for me that the line above "tue-get install tue-dev" isn't actually needed. If that line is required, maybe change it to "# also install extra tools" ?

MatthijsBurgh:
see #159

alberth:
```` shell
tue-get install tue-dev
tue-get install tue-dev-full #tue-dev plus extra tools
````
#159 now really makes that the latter includes the former, as I thought it would be the case.
That makes the first line obsolete, doesn't it?

alberth:
@MatthijsBurgh  close enough, thanks

PR #176, opened 2017-10-10
Add wmctrl to dashboard target

MatthijsBurgh:


Rayman:
It's part of the alias, not of the package

PR #181, opened 2017-10-17
Update tue-status

MatthijsBurgh:


LoyVanBeek:
I'm not deep enough into the tue-env tools so I'm not the guy to review this.

PR #183, opened 2017-10-19
broken dependency

MatthijsBurgh:
```
from html5lib import treebuilders, inputstream
ImportError: cannot import name inputstream
```
https://github.com/xhtml2pdf/xhtml2pdf/issues/318

### DIFF
--- a/installer/targets/ros-python-xhtml2pdf/install.yaml
+++ b/installer/targets/ros-python-xhtml2pdf/install.yaml
@@ -1,2 +1,4 @@
 - type: pip
   name: xhtml2pdf
+- type: pip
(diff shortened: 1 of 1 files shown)


Rayman:
If you update  `xhtml2pdf` to the latest version, this is already fixed

MatthijsBurgh:
@Rayman upgrade to latest version still gives xhtml2pdf version 0.0.6 and a broken dependency to html5lib

Rayman:
Ah yeah you're right. I tried `import xhtml2pdf ` which succeeded, but the following failed:
```python
from xhtml2pdf import pisa
```

PR #192, opened 2017-10-30
Update matthijs target

MatthijsBurgh:


installer/targets/matthijs/setup
```diff
@@ -10,5 +10,10 @@ function sb
 source ~/.bashrc
 }
 
+function gitpull
```
LoyVanBeek: This is a config option of git, no need to use an alias for that

Rayman: https://git-scm.com/docs/git-config#git-config-pullff

MatthijsBurgh: What would be the best way to apply it? By adding 'git config --global --add pull.ff only' to install.bash? I don't think you would need a check to prevent execution each install, because it really doesn't take any time. Do you agree?

MatthijsBurgh:
@LoyVanBeek please approve again

installer/targets/matthijs/setup
Rayman: Don't run `git config` in your bashrc. I did that once and when you open multiple terminals it will do wierd things.

MatthijsBurgh: No, I added to the install script. Because setting it in each new terminal is also crazy.

PR #194, opened 2017-10-31
Feature/chrony local sync

reinzor:
Test according to http://linux-hacking-guide.blogspot.nl/2015/05/configure-ntp-using-chrony-for-isolated.html

we can sync manually using:

    amigo-sync-ubuntu-ntp


reinzor:
Should solve https://github.com/tue-robotics/tue_robocup/issues/461#event-1319546587

Timple:
This does throw away some of the evolution in the older files. 
For example: empty motherboard batteries are not supported anymore (was covered by amigo1 syncing with the other amigo PCs on boot only).
Also amigo will always be off by a lot I guess. Who's gonna type the alias and when?

Also, are the hardware clocks written periodically by the default chrony config (covers power failure / hard resets)?

PR #196, opened 2017-11-04
Add python script to check ssh connection, fixes #157

### LINKED ISSUE #157: sshamigo1 gives no output if command fails
If a new amigo PC is installed (or for other reasons) the keys change.

Then sshamigo2 does not work anymore, ssh amigo@amigo2 also does not work anymore but at least gives some useful output.

Proposal: also show output on such errors

Timple:
Current alias masks errors originating from wrong keys. Happens on replacing amigo computers

Fixes #157


MatthijsBurgh:
Can connect to AMIGO. But not sure if it does what it should do. Trusting you on that.

Timple:
I'll merge right before I'm there again for support ;-)

installer/targets/networking/install.bash
```diff
@@ -1,2 +1,3 @@
 # Install config file (mdns4_minimal is normally missing and should be present)
 tue-install-cp nsswitch.conf /etc/nsswitch.conf
+tue-install-cp wait-for-host.py /usr/bin/wait-for-host.py
```
Rayman: If you are doing this, consider creating a package

Timple:
Fair enough, I'll look into this.

Timple:
Sure, I'm still using it actively at work. But as Rayman said, some troubleshooting commands so people can help themselves would be wiser

PR #198, opened 2017-11-05
Multiple fixes/features

MatthijsBurgh:


setup/tue.bash_functions
```diff
@@ -225,7 +225,7 @@ function _tue-repo-status
         if [ $? -eq 0 ]
         then
             # Is git
-            if echo "$res" | grep -q '\['   # Check if ahead of branch
+            if echo "$res" | grep -q 'behind' ||  echo "$res" | grep -q 'ahead' # Check if behind or ahead of branch
```
Timple: This is what regular expressions are for ;-)
if echo "$res" | grep -q -E 'behind|ahead' # Check if behind or ahead of branch

LoyVanBeek: But this is quite understandable, more than the average regex :-)

PR #199, opened 2017-11-06
remove install.bash, because also install.yaml available

MatthijsBurgh:


LoyVanBeek:
LGTM. @reinzor why was this added like this at the time ([3yrs ago](https://github.com/tue-robotics/tue-env/blame/a1ddb6b8f6b5d05dabfe0d228191e244d913eec7/installer/targets/ros-rosapi/install.bash))

MatthijsBurgh:
The yaml filed was added a year ago, so it has been 'double installed' for a long time

PR #220, opened 2018-01-09
added target gpsr-cmdgen and corresponding aliases

timassman:
Added a tue-env target to install the command generator from robocup.
I tried to do this with an install.yaml instead of install.bash, but I only saw examples for git clones of ros repositories. The git clones for non-ros repositories (openface, openpose) also used bash scripts like this. But I think the git repository will not be automatically updated this way, when commands are updated by the committee?

please test by:

1) checking out branch from Matthijs
2) (not sure if needed: tue-get update)
3) tue-get install cmdgen
4) cmdgen-spr
5) cmdgen-gpsr
6) cmdgen-eegpsr 

if questions are asked "file exist, overwrite?" I chose "no"

If this works, I will update the README in the corresponding challenges


installer/targets/gpsr-cmdgen/install.bash
```diff
@@ -0,0 +1,17 @@
+if [ ! -d ~/GPSRCmdGen ]; then
+	pushd .
```
MatthijsBurgh: Not needed, this is taken care of by the installer.

installer/targets/gpsr-cmdgen/setup
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+alias amigo-cmdgen-spr="mono ~/GPSRCmdGen/bin/Release/SPRTest.exe"
```
MatthijsBurgh: Why are these aliases prefixed with AMIGO? These are generic for all robots.

installer/targets/gpsr-cmdgen/install.bash
```diff
@@ -0,0 +1,17 @@
...
+
+	echo "Installing GPSRCmdGen"
+
+	echo "Required for GPSRCmdGen: mono-complete"
+	sudo apt-get install mono-complete
+
+	echo "Cloning GPSRCmdGen from git repository"
+	git clone http://github.com/kyordhel/GPSRCmdGen.git ~/GPSRCmdGen --recursive
```
MatthijsBurgh: Please use ~/src/XXX

PR #228, opened 2018-01-22
Adding assume-yes to cmdgen install

MatthijsBurgh:
This is the reason tue_robocup PR is failing


installer/targets/cmdgen/install.bash
```diff
@@ -6,7 +6,7 @@ then
     echo "Installing GPSRCmdGen to ~/src/GPSRCmdGen"
 
     echo "Installing dependency of GPSRCmdGen: mono-complete"
-    sudo apt-get install mono-complete
```
Timple: Why not use:
tue-install-system-now

MatthijsBurgh: Merging for now. Changes can applied when other PR is merged.

PR #229, opened 2018-01-22
Nobleo-backport

Timple:
With:
Support for branches: Usefull for CI
More robust package installing options
Install snapd if using snaps
Activate docker permissions on install
Install less packages initially when installing ROS


installer/scripts/tue-install-impl
```diff
@@ -223,7 +223,25 @@ function tue-install-git
             cres=
         fi
 
-        res=$res$cres
+        res="$res $cres"
+        cd $mem_pwd
+    fi
+
+    if [ -n "$BRANCH" ]; #Cannot be combined with version-if because this one might not exist
```
MatthijsBurgh: Where is BRANCH defined? I don't see any difference to version, except from the error checking.

installer/scripts/tue-install-impl
```diff
@@ -363,6 +381,35 @@ function tue-install-system
...
+    local install_required=false
+
+    for pkg in "$@"
+    do
+        # Check if pkg is not already installed dpkg -S does not cover previously removed packages
+        # Based on https://stackoverflow.com/questions/1298066
+        if ! dpkg-query -W -f='${Status}' $pkg 2>/dev/null | grep -q "ok installed"
+        then
+            tue-install-debug "Package $pkg requires installation"
+            install_required=true
```
MatthijsBurgh: It would be nicer to create a list of pkgs which are not installed yet. And then check if the list is empty. That is also the way it is done in tue-install-system.

installer/scripts/tue-install-impl
```diff
@@ -732,6 +779,8 @@ if [ -n "$TUE_INSTALL_SNAPS" ]; then
 
     TUE_INSTALL_CURRENT_TARGET="SNAP"
 
+    _make_sure_installed snapd
```
MatthijsBurgh: I would add this to the _make_sure_installed on top of this script

installer/scripts/tue-install-impl
```diff
@@ -34,7 +34,7 @@ function _make_sure_installed
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
 # Make sure tools used by this installer are installed
-_make_sure_installed python-yaml git subversion python-pip
+_make_sure_installed python-yaml git subversion python-pip snapd
```
Timple: Ik had bewust snapd pas geinstalleerd als je ook snaps installeerd. Is best wel een zware dependency anders. Die ook bij elke CI etc aangeroepen wordt.

installer/scripts/tue-install-impl
```diff
@@ -381,32 +381,32 @@ function tue-install-system
...
     do
         # Check if pkg is not already installed dpkg -S does not cover previously removed packages
         # Based on https://stackoverflow.com/questions/1298066
         if ! dpkg-query -W -f='${Status}' $pkg 2>/dev/null | grep -q "ok installed"
         then
-            tue-install-debug "Package $pkg requires installation"
-            install_required=true
-        else
-            tue-install-debug "Package $pkg is already installed"
+            pkgs_to_install="$pkgs_to_install $pkg"
```
Timple: Owja, dat is natuurlijk mooier!

installer/scripts/tue-install-impl
```diff
@@ -223,7 +223,25 @@ function tue-install-git
...
+    if [ -n "$BRANCH" ]; #Cannot be combined with version-if because this one might not exist
+    then
+        mem_pwd=$PWD
+        cd $2
+
+        tue-install-debug "cd $2 && git checkout $BRANCH"
+
+        cres=$(git checkout $BRANCH 2>&1)
+        tue-install-debug "$cres"
+        if [[ $cres == "Already on "* || $cres == "error: pathspec"* ]]
```
MatthijsBurgh: good to apply this error catching also to other 'git checkout' in the installer

installer/targets/ros/install.bash
```diff
@@ -8,15 +8,15 @@ if [ ! -d /opt/ros/$TUE_ROS_DISTRO ]
...
+
     sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
 
     wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
 
-    sudo apt-get update
+    sudo apt-get update -qq
 
     # Install basic ROS packages. All other packages will be installed using tue-rosdep
-    sudo apt-get install  --assume-yes ros-$TUE_ROS_DISTRO-ros-base cmake python-catkin-pkg python-empy python-nose python-setuptools libgtest-dev build-essential python-catkin-tools
```
MatthijsBurgh: Are you sure that we don't need python-setuptools for the setup.py files? And python-nose for test?

MatthijsBurgh:
@Timple, please implement last comments

installer/targets/ros/install.bash
MatthijsBurgh: python-setuptools and python-nose are dependencies of the rest, so these will be installed. I don't know if replacing ros-base with ros is ok. We need to test that with a clean install.

installer/targets/ros/install.bash
```diff
@@ -31,7 +33,7 @@ TUE_DEV_DIR=$TUE_ENV_DIR/dev
 if [ ! -f $TUE_SYSTEM_DIR/devel/setup.bash ]
 then
     mkdir -p $TUE_SYSTEM_DIR/src
-    hash g++ 2> /dev/null || sudo apt-get install --assume-yes g++
+    hash g++ 2> /dev/null || tue-install-system-now g++
```
MatthijsBurgh: @Timple We don't need the hash check here anymore, right? But it is faster than the dpkg check right? Than it is usefull to keep these.

Timple:
I don't mind either way. tue-install-system-now is more descriptive and
consistent.

Hash is 0.07s faster on my PC.

2018-03-25 11:15 GMT+02:00 Matthijs van der Burgh <notifications@github.com>
:

installer/targets/ros/install.bash
Timple: I don't mind either way. tue-install-system-now is more descriptive and
consistent.

Hash is 0.07s faster on my PC.

MatthijsBurgh:
@Timple, please check one more time before merging

PR #231, opened 2018-02-03
Use our fork of the ROS diagnostic_updater

LoyVanBeek:
Forked at https://github.com/tue-robotics/diagnostics

Used in https://github.com/tue-robotics/image_recognition/pull/55.

### DIFF
--- a/installer/targets/ros-diagnostic_updater/install.yaml
+++ b/installer/targets/ros-diagnostic_updater/install.yaml
@@ -1,5 +1,5 @@
 - type: ros
   source:
     type: git
-    url: https://github.com/tue-robotics/diagnostics
+    url: https://github.com/tue-robotics/diagnostics.git
     sub-dir: diagnostic_updater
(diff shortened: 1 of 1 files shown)


LoyVanBeek:
@MatthijsBurgh Can you verify that installing this works OK? My ROS env is not a clean tue-env and never has been. Getting this version with that setup was a hassle but it shouldn't be with a normal tue-env.

MatthijsBurgh:
Jep, installing is working. Can be build and rosrun works.

Timple:
I am missing the rationale here.

Why is ours better?

LoyVanBeek:
Fair points! We want to have some diagnostics on the Jetson board, to see if it is up and running.
I was making a diagnostic_updater that publishes a heartbeat but since this was very generic, @reinzor suggested to add this to the standard ROS diagnostics, so I did. 

I forked it and submitted a PR at https://github.com/ros/diagnostics/pull/73

PR #233, opened 2018-02-06
Use local hostname resolving to get rid of IP ranges.

Timple:
Lets evaluate this on amigo for some challenges


setup/tue.bash_functions
```diff
@@ -803,7 +803,7 @@ function tue-robocup-update
             cd $pkg_dir
             local current_url=`git config --get remote.origin.url`
 
-            if echo "$current_url" | grep -q "192.168.44.10"
+            if echo "$current_url" | grep -q "froboticssrv.local"
```
LoyVanBeek: typo?

LoyVanBeek:
Who does the DNS resolving now? The router? Do we have the router config back-upped somewhere?

setup/tue.bash_functions
Timple: Fixed

Timple:
DNS is now done with /etc/hosts file. Router is not involved. So no need for a backup (regarding this PR)

PR #234, opened 2018-02-06
Add ros-matrix_creator_hal target

Timple:


reinzor:
Please test then before approving :)

PR #235, opened 2018-02-20
Local hostname resolving

MaxBaeten:
Added some small bugfixes.


installer/targets/amigo1/chrony.conf
```diff
@@ -5,7 +5,7 @@ allow 192.168/16
 local stratum 8
 
 # All offsets over 1 second are corrected immediately at start. Each of the amigos is rapidly polled several times, and a majority voting mechanism used to find the most likely range of system clock error that is present.
-initstepslew 1 amigo2.local amigo3.local amigo4.local
+initstepslew 1 amigo2.local amigo3.local tegra-ubuntu.local
```
MatthijsBurgh: I think it should be renamed to amigo4.

PR #240, opened 2018-03-03
able to provide max depth of dependencies

MatthijsBurgh:


Timple:
Looks good, let me test it first. What is the reason you want this?

MatthijsBurgh:
In the description it said that level was an option, which wasn't implemented. So I did :p

PR #257, opened 2018-05-02
Check against echo, print original res

MatthijsBurgh:
This keeps the style of the git output. While not printing, if msg is empty.


Timple:
That seems to be the case indeed. Weird...

PR #261, opened 2018-05-12
Include ros-vizbox in robocup-at-home

LoyVanBeek:
That is a dependency of tue-dev so you should get the vizbox if all is correct

### DIFF
--- a/installer/targets/ros-python-pillow/install.yaml
+++ b/installer/targets/ros-python-pillow/install.yaml
@@ -1,0 +1,2 @@
+- type: pip
+  name: pillow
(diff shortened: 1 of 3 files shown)


installer/targets/ros-python-pillow/install.yaml
```diff
@@ -0,0 +1,2 @@
+- type: pip
+  name: pillow
```
MatthijsBurgh: Is there also a debian target?

installer/targets/ros-python-tornado/install.yaml
```diff
@@ -0,0 +1,2 @@
+- type: pip
+  name: tornado
```
MatthijsBurgh: Is there also a debian target?

MatthijsBurgh:
@LoyVanBeek do you aprrove my changes? Than I will merge it.

LoyVanBeek:
I just merged your PR in the vibox, so this is good to go as welll

PR #262, opened 2018-05-14
Implement general git-hook structure for all repos and add git credential checker if username is amigo.

Timple:
Solves tue-robotics/tue_robocup#599

Please note that after an `tue-get install git-hooks` a `sudo apt install git` is required. This will upgrade git itself.

I am not going to enforce this, because
* it is quite a general ppa issue that could potentially be solved by the tooling
* it does not break backwards compatibility (its not working now either)
* it is only required on the amigo PC's as long as we only have the 'author' check.


installer/targets/git-hooks/git_hooks/pre-commit.d/author
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+AUTHORINFO=$(git var GIT_AUTHOR_IDENT) || return 1
+NAME=$(printf '%s\n' "${AUTHORINFO}" | sed -n 's/^\(.*\) <.*$/\1/p')
+EMAIL=$(printf '%s\n' "${AUTHORINFO}" | sed -n 's/^.* <\(.*\)> .*$/\1/p')
+
+echo "I'm going to commit with \"$NAME <$EMAIL>\""
+if [[ "${NAME}" == "amigo" ]]
```
MatthijsBurgh: A lot of athome laptop and also personal laptops use amigo as username. So this check will make me and a lot of people angry. So please convert to hostname or use the environment variable ROBOT_REAL.

installer/targets/git/install.bash
```diff
@@ -9,6 +9,7 @@ if [ ! -f /usr/local/bin/git-extras ]; then
 	git config --global color.ui true
 fi
 
+tue-install-ppa ppa:git-core/ppa #Git 2.9 required for git-hooks
```
MatthijsBurgh: Why are these not in a yaml?

installer/targets/git-hooks/git_hooks/pre-commit.d/author
```diff
@@ -0,0 +1,26 @@
...
+then
+    echo "Please commit with a username, run \"git su <name>\" to swith users"
+    return 1
+fi
+
+if [[ "${USER}" == "amigo" ]]
```
MatthijsBurgh: Change username check here too

installer/targets/git-hooks/git_hooks/pre-commit.d/author
Timple: For the athome laptops this check still makes sense no?

But for the personal ones it does not...

Timple:
I think it is OK now.

The `[[ "${NAME}" == "amigo" ]]` is still in place, because amigo as commit-username is still not allowed. Not even on your own laptop ;-)

MatthijsBurgh:
Both the username and robot_real checks have been tested. 

Issues:
As you said, git needs to be updated manually. A version check and auto update would be nice.
When the username consists capital letters, the name check isn't working correctly
(I would prefer if the printing of the name, still will be done with capital letters)

installer/targets/git-hooks/git_hooks/pre-commit.d/author
```diff
@@ -0,0 +1,26 @@
...
+then
+    # prompt for conformation
+    exec < /dev/tty
+    read -p "Are you sure? " -n 1 -r
+    exec <&-
+    echo    # (optional) move to a new line
+    if [[ ! $REPLY =~ ^[Yy]$ ]]
+    then
+        echo "You can change the user by running \"git su <name>\""
+        exit 1
```
MatthijsBurgh: Any reason for using 'exit' here and 'return' in the other check?

Timple: Eeehm, nope.

MatthijsBurgh: So, please use the same. Up to you which is the best.

PR #263, opened 2018-05-15
replace .tue by $TUE_DIR, when possible

MatthijsBurgh:
To reduce dependency of having the installer in ~/.tue


MatthijsBurgh:
@Timple, could you review this one as well?

PR #267, opened 2018-06-08
new git workflow

MatthijsBurgh:
for all the scripts on the server side:
`~/MEGA/tools/server`

please review as soon as possible.

### DIFF
--- a/setup/tue.bash_functions
+++ b/setup/tue.bash_functions
@@ -443,7 +443,7 @@ function tue-get
     local tue_dep_dir=$TUE_ENV_DIR/.env/dependencies
     local tue_installed_dir=$TUE_ENV_DIR/.env/installed
 
-    cmd=$1
+    local cmd=$1
     shift
 
     #Create btrfs snapshot if possible and usefull:
@@ -461,7 +461,7 @@ function tue-get
 
     if [[ $cmd == "install" ]]
     then
-        $TUE_DIR/installer/scripts/tue-install $@
+        $TUE_DIR/installer/scripts/tue-install $cmd $@
         error_code=$?
 
         [ $error_code -eq 0 ] && source ~/.bashrc
@@ -484,7 +484,7 @@ function tue-get
 
         if [ $error_code -eq 0 ]
         then
-            $TUE_DIR/installer/scripts/tue-install $@
+            $TUE_DIR/installer/scripts/tue-install $cmd $@
             error_code=$?
             [ $error_code -eq 0 ] && source ~/.bashrc
         fi
--- a/setup/tue.bash_functions
+++ b/setup/tue.bash_functions
@@ -254,7 +254,7 @@ function _tue-repo-status
             fi
 
             local current_branch=`git rev-parse --abbrev-ref HEAD`
-            if [ $current_branch != "master" ] && [ $current_branch != "develop" ] && [ $current_branch != "indigo-devel" ] && [ $current_branch != "kinetic-devel" ] && [ $current_branch != "toolchain-2.9" ]
+            if [ $current_branch != "master" ] && [ $current_branch != "robocup" ] && [ $current_branch != "develop" ] && [ $current_branch != "indigo-devel" ] && [ $current_branch != "kinetic-devel" ] && [ $current_branch != "toolchain-2.9" ]
             then
                 echo -e "\033[1m$name\033[0m is on branch '$current_branch'"
             fi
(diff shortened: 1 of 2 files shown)


setup/tue.bash_functions
```diff
@@ -722,23 +851,79 @@ function tue-robocup-set-timezone-home
...
     # If directory already exists, return
-    [ -d $pkg_dir ] && return
+    [ -d $repo_dir ] && return
+
+    # If there is internet, use it for first clone. Otherwise an error
+    # will be raised during switching back to default branch of origin
+    local internet=$(echo _ping_bool github.com)
+    if $internet
+    then
+        git clone https://github.com/tue-robotics/${1}.git $repo_dir
```
Rayman: I don't like different behavior based on whether or not internet is available. Can you make it that internet is required? You can always do a clone in the hotel and most of the time internet is available.

Rayman:
The idea behind the PR is good btw

setup/tue.bash_functions
MatthijsBurgh: The behind this function (in my mind) is:
installing a package without internet available and also setting it up correctly, so that is works correctly with all the other tools.

If you have internet you could just use tue-get, but than it is not setup to use the server.  The current tool to switch to the server, shouldn't be used during a tournament. Because it checks out the robocup branch for all repos, which is bad if you are actually on an other branch at that moment.

I could change this to only checking out robocup, if you are on a branch which has github as a remote. Then you could run the tool to switch to the server also during tournament.

Then the tue-robocup-install-package is only for situations without internet. This has only one issue with the tool for switching back to github. That tool tries to checkout the symbolic ref `refs/remotes/origin/HEAD`. This is only created for the remote used at cloning. So if you clone from the server, rename that one to roboticssrv. Then add github as origin, the symbolic ref doesn't exist for origin. (This is also a problem right now and the reason for the different behaviour, because you would like to avoid that situation.). If I change the function that it will checkout master if no symbolic ref exists, that is fixed.

For with internet you then could you use:
tue-get install TARGET
tue-robocup-set-server

For without internet:
tue-robocup-install-package PKG

LoyVanBeek:
Nice work!

So, what is the workflow and list of commands each of us should run pre- and  post-robocup?

MatthijsBurgh:
Pre:
tue-robocup-set-server

Post:
tue-robocup-set-github

On the server side (scripts right now, maybe I will move these to tue-env as well instead of MEGA):
Pre:
mirror-github.bash TAG(i.e. 2018-before-rwc)

Post:
push-mirror.bash BRANCH(i.e. rwc2018)

LoyVanBeek:
OK, but that means that the branch name you work with locally during RoboCup is not the same branch name as you'll have post-robocup. 

I'd say: check out a branch 'rwc2018' with roboticssrv.local as the remote and push that to github with the same name. Otherwise, there will be confusion and huge conflicts and mess I expect. 

If the first command is ```tue-robocup-set-server``` and then eg. ```tue-robocup-checkout BRANCH rwc2018``` or something, that would solve this already

MatthijsBurgh:
On the tournament you will have robocup(roboticssrv). After the tournament this branch is pushed to github as (i.e. rwc2018). All laptops have to be set back to github branches. Then during the PR process, you check out rwc2018(github) on your laptop. This prevents having two branches with the same name, but different remotes. Which could be more confusing in my opinion.

LoyVanBeek:
> This prevents having two branches with the same name, but different remotes.
They are the same branch but with different remotes. Git is made to do this sort of stuff. 
Your workstation is also just a remote, the only difference between remotes is how many of the commits of a branch they have.

If you pull the latest changes from roboticssrv and directly push them to github, there is no need to push from roboticssrv to github, it's just the same commits.

> All laptops have to be set back to github branches

Branches and remotes remotes are different things, but all remotes can mirror the same branches. They can go out of sync with each other but that is what push/pull are for. 
All laptop will have to pull that same branch from a different remote. eg. ```git pull origin rwc2018``` when pulling from github or  ```git pull roboticssrv rwc2018```. 

To set the deafult: ```git branch --set-upstream-to=origin/rwc2018``` to go pull from github by default again.

MatthijsBurgh:
To keep it simple and robust, the branch on the tournament should always be `robocup` and hardcoded in the code. To prevent typos etc.

I prefer to have rwc2018 on github, because it is much nicer in the PRs. This is a point for discussion. Otherwise we can also use `robocup` on github,

PR #270, opened 2018-07-16
don't force update libcppunit-dev anymore

MatthijsBurgh:
all pc's should have been updated by now


Timple:
What about new pc's?

MatthijsBurgh:
These will automatically get the new version. This fix was needed, because we don't update if already installed.

PR #271, opened 2018-07-16
Install ccache also on robots

MatthijsBurgh:
small refactor in robot and robot-user target

### DIFF
--- a/installer/targets/ccache/install.yaml
+++ b/installer/targets/ccache/install.yaml
@@ -1,0 +1,2 @@
+- type: system
+  name: ccache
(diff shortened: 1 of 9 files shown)


Timple:
I would just install robot-user on the robots. 
Because if we wouldn't develop, we wouldn't need to install gitsu etc also.

robot-user-robot-target is a bit unclear.

MatthijsBurgh:
I don't want to install robot-user on robots, because than we need to refactor it back, when we want to add a target for robot-users only. I will revert the common target. So `robot` and `robot-user` will be nearly identical again.

PR #283, opened 2018-10-09
Add global gitignore

MatthijsBurgh:
In my opinion this is best way to automate the configuration. This way it will be overwritten everytime the installer is runned, but I wasn't able to find easy check.


Timple:
I've quickly backported our solution. Pick your flavor ;-)

PR #284, opened 2018-10-12
Backport gitignore file if none exists

Timple:
Similar to #283 

@MatthijsBurgh Let me know which you prefer, this is what we have at Nobleo.

If no global gitignore file has been configured yet (as advanced users might). A default one is used with most used rubish files.


MatthijsBurgh:
@Timple please review my changes.

PR #291, opened 2018-10-18
inactive color offset = 1 in terminator config

MatthijsBurgh:


MatthijsBurgh:
This only has influence when installing for the first time.

PR #292, opened 2018-10-18
Fixes in tue-env

ar13pit:
Executing `tue-env config` would first create a new directory user_setup.bash in the current directory. Which made no sense so removed it.

Question for @MatthijsBurgh 
`tue-env remove ENV` only removes the environment file from `$TUE_DIR/user/envs` but not the `ENV_DIR`. Should that be removed as well or not?


MatthijsBurgh:
I will remove myself as reviewer. Please add me again, when all the work is done.

MatthijsBurgh:
I am ok with keeping the directory. When creating a new one with the same name, The old one should be suffixed with date and time.

ar13pit:
@MatthijsBurgh I am done making changes. About the removal of directories, I have made it as a complete remove as `tue-get remove` doesn't do that. If you want I could add a user confirmation statement there.

MatthijsBurgh:
I don't prefer the complete removal of the environment. As I suggested before, keep it. And rename it, in case a new environment is created with the same dir. @Timple What do you prefer?

ar13pit:
With renaming I see just one flaw that the user has no clue about what's happening on his computer and ends up storing the directories.

My suggestion is to include a [Y/N] question about removal of the directory. If the user says "Y" then remove it else rename it.

Timple:
Never worked with these environments. The only thing I know about them is that they don't work if you interrupt fresh installation.
So I don't have an opinion about this.

MatthijsBurgh:
@ar13pit could you implement the things we discussed.

ar13pit:
@MatthijsBurgh Its done and I have tested it as well.

ar13pit:
@MatthijsBurgh review and let me know if the structure is good enough now. I have tested it and it works. In my opinion if you want to do another level of parameter handling, it will get ugly. 

It can be cleanly done by defining functions and calling them in the switch case but that will export all such functions into the bash environment because of `setup.bash`.

MatthijsBurgh:
@ar13pit you moved to implementation into the parameter handling. What I suggested is that you first do the parameter handling. In which you set some variables. After which you have some if statements, which check the variables which are set in the parameter handling.

ar13pit:
I see. I think I understood it differently. I'll fix it :laughing:

MatthijsBurgh:
@ar13pit ping ;)

ar13pit:
@MatthijsBurgh I think this is done now. What do you think?

PR #296, opened 2018-11-10
convert spotify to snap target

MatthijsBurgh:


MatthijsBurgh:
Music is very critical :p

PR #306, opened 2018-11-29
Fix pip upgrade

MatthijsBurgh:
Fixes:
- Version checking was wrong
- user version was updated/installed, while system version is used
- remove unused if check, if statement was the same as if statement of higher level

Before fix: https://travis-ci.org/tue-robotics/telegram_ros/builds/461249039#L2710
After fix: https://travis-ci.org/tue-robotics/telegram_ros/jobs/461252304#L2803

### DIFF
--- a/installer/scripts/tue-install-impl
+++ b/installer/scripts/tue-install-impl
@@ -776,10 +776,11 @@ if [ -n "$TUE_INSTALL_PIPS" ]; then
 
     if [[ ! $(pip --version | awk '{print $2}') == '9.'* ]]; then
         tue-install-debug "pip not yet version >=9, but `pip --version | awk '{print $2}'`"
-        pip install --user --upgrade pip
+        sudo -H pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
No more pip main error in the CI

alberth:
Not failing in one particular case versus correct at all times are very different notions. In particular, the former does not imply the latter.

MatthijsBurgh:
> Not failing in one particular case versus correct at all times are very different notions. In particular, the former does not imply the latter.

True, but it has been failing in all the CI. Everywhere the same script is used. So if one is fixed, probably all are fixed.

alberth:
So you're using this one particular case a lot. As computers tend to be quite good at doing the exact same thing every time, yes, it's likely they all fail or they all work.
However it's still only one case.

While correct implies working, there is no such relation from working to correct. In fact, I have seen working solutions that are just plain wrong.

PR #312, opened 2018-12-06
update rosdep once every powercycle

MatthijsBurgh:
update rosdep, otherwise it can cause issues with catkin_lint


ar13pit:
@MatthijsBurgh How does this run every powercycle? To my knowledge install.bash runs only when either tue-get install or update commands are issued.

MatthijsBurgh:
Maybe a bit misleading. But the file `/tmp/rosdep_update` is created after running `rosdep update`. If the file does exist, the if returns false, so it isn't updated anymore. Until you reboot and /tmp is cleared.

PR #313, opened 2018-12-06
correct target name for python-graphviz-pip, matching rosdistro

MatthijsBurgh:


ar13pit:
@MatthijsBurgh How does catkin find python-graphviz-pip when there is no package of that name?

ar13pit:
For reference:
`python-graphviz-pip` comes from [here](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L1622) in rosdistro

PR #314, opened 2018-12-07
add ruby-dev

MatthijsBurgh:
Was needed


MatthijsBurgh:
Because all dependencies should be there as a `ros-X` target

ar13pit:
I guess the next improvement to tue-env should be this. Have only ros packages with target names starting with `ros-`. It would just be cleaner in my opinion.

MatthijsBurgh:
It is better, no prio at the moment

PR #318, opened 2018-12-15
Repository cleanup

ar13pit:
Removed the `create` and `make` directories as the scripts there are no longer in use.

Also where is `.editorconfig` being used and why is it there in the repo root?


ar13pit:
Because there are some scripts from `make` directory there and `catkin_make` stuff in the switch case. I thought that its not being used. Maybe I should just get rid of those?

MatthijsBurgh:
Make scripts aren't used anymore. And the case in the tue-make function is there, because some people like to use catkin_make.

ar13pit:
@MatthijsBurgh Need approval to merge.

PR #320, opened 2018-12-23
Updating README

ar13pit:
Added how to create a target in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,4 +46,28 @@ Name of the target must start with `ros-` only if it will be compiled/installed 
 
 ### Writing `install.yaml`
 The following combinations are possible:
+1. ROS package install
+  a. From source
+```yaml
+- type: ros [ros-rosbuild]
+  source:
+    type: git [svn]
+    url: <Repository URL>
+    sub-dir: <Sub directory of the repository> (Optional field)
+    version: <Version to be installed> (Optional field)
+```
+  b. From system
+```yaml
+- type: ros
+  source:
+    type: system
+    name: <Package name>
+```
+
+2. Target / System / PIP / PPA / Snap
+```yaml
+- type: [target/system/pip/ppa/snap]
+  name: <Name of the candidate>
+```
+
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,3 +29,53 @@ command in a shell.
...
+2. `install.bash`
+3. `setup`
+
+Installation happens in the above order. First `install.yaml` is parsed and the
+instructions in it are executed. Then `install.bash` is executed. This must have
+commands/instructions that cannot be specified in the YAML file. Lastly, the
+`setup` file is sourced in the bash environment by `setup.bash` of tue-env.
+
+### Naming conventions
+Name of the target must start with `ros-` only if it will be compiled/installed by ROS.
```
MatthijsBurgh: Must start with `ros-` if target is installed by reading the dependencies from a catkin `package.xml`.

README.md
```diff
@@ -29,3 +29,53 @@ command in a shell.
 ```bash
 echo -e "export TUE_ENV=$TUE_ENV\nsource ~/.tue/setup.bash" >> ~/.bashrc
 ```
+
+## Create a new target
+A target must consist of the following three files:
```
MatthijsBurgh: not must, **can**

README.md
ar13pit: This was a bit unclear to me previously as well. Lets say we want to create a target A which will be installed with System/PIP/PPA/Snap then why name it with `ros-`? This target by itself has no `package.xml` but it acts as a dependency to a catkin package. In my opinion if this target had `package.xml` or `roscd` could find it, only then we should name it with `ros-` else it can be misleading.

What do you think about this?

MatthijsBurgh: That is true. The current implementation is misleading. But suggested implementation is difficult to implement. Because the prefix is not depending on the target itself, but on parent dependencies. So I suggest to keep the current implementation or remove the prefix at all.

ar13pit: For now I'll keep the current setting. But I am opening a new issue where in I'll address the naming conventions and fix them.

PR #323, opened 2018-12-23
Fix names of scripts with some minor updates

### LINKED ISSUE #321: Update naming conventions for scripts
The scripts in the repository have no naming convention. Bash scripts must end with `.bash` and python scripts must end with `.py`

ar13pit:
Fixes #321


MatthijsBurgh:
Before you are doing too much work. There are some naming conventions for both bash and python for when to use extension and not. As far as I know bash doesn’t use the extension to determine how to run/open a file. Therefore you need a shebang line. (But not when a file is only sourced, I read that somewhere on stackoverflow)

ar13pit:
I agree. The only reason why I am changing extensions is to make files easier to identify. Execution wise it makes no sense to change their extensions.

ar13pit:
You mean the bootstrap script right?

MatthijsBurgh:
Yep

ar13pit:
I have tested it out on VirtualBox now and it works fine. Only thing is for a fresh installation `catkin: command not found` will be seen after the `wget` statement as ROS or catkin tools won't be installed and `setup.bash` has a `catkin` call as the last statement. Apart from that it works perfectly fine.

installer/scripts/bootstrap.bash
```diff
@@ -38,6 +38,7 @@ fi
 
 # Source the installer commands
 source ~/.tue/setup.bash
+echo "[bootstrap] Loading tue-env"
```
MatthijsBurgh: This hardly takes any time, so unnecessary in my opinion. And if you put it there, put it before sourcing it.

installer/scripts/bootstrap.bash
```diff
@@ -1,30 +1,43 @@
 #!/bin/bash
```
MatthijsBurgh: Why not check ubuntu version and use that to determine the ros name. Then you don't need an argument and you can use the old source+wget structure.

installer/scripts/bootstrap.bash
```diff
@@ -1,30 +1,43 @@
...
-    git clone https://github.com/tue-robotics/tue-env.git ~/.tue
+    git clone -b ar13pit-patch/fix-names-scripts-targets https://github.com/tue-robotics/tue-env.git ~/.tue
 fi
 
 # Source the installer commands
 source ~/.tue/setup.bash
-
-# Set ROS version
-TUE_ROS_DISTRO=kinetic
+echo "[bootstrap] Loading tue-env"
```
MatthijsBurgh: This hardly takes any time, so unnecessary in my opinion. And if you put it there, put it before sourcing it.

installer/scripts/bootstrap.bash
ar13pit: That's a good idea. I'll add that.

installer/scripts/bootstrap.bash
```diff
@@ -1,31 +1,54 @@
...
     done
-    mv -f ~/.tue ~/.tue.$(date +%F_%R)
+    mv -f ~/.tue ~/.tue.$date_now
 fi
 
 if [[ -z "$CI" ]]
 then
     # Update installer
-    git clone https://github.com/tue-robotics/tue-env.git ~/.tue
+    git clone -b ar13pit-patch/fix-names-scripts-targets https://github.com/tue-robotics/tue-env.git ~/.tue
```
MatthijsBurgh: Neede to be reverted

MatthijsBurgh:
Doesn't really fix #321. I think this one is ready to be merged.

ar13pit:
I agree. Hence I have updated #321 and opened up #328 to continue working on what was started.

PR #334, opened 2018-12-28
Create tue-misc.bash for everything which doesn't fit somewhere else.

MatthijsBurgh:


setup/tue-misc.bash
```diff
@@ -4,11 +4,13 @@ export LC_ALL="C.UTF-8"
...
 # pip bash completion
-
-_pip_completion()
-{
-    COMPREPLY=( $( COMP_WORDS="${COMP_WORDS[*]}" \
-                   COMP_CWORD=$COMP_CWORD \
-                   PIP_AUTO_COMPLETE=1 $1 ) )
-}
-complete -o default -F _pip_completion pip
+if hash pip 2> /dev/null
```
ar13pit: Wouldn't it be better to have pip explicitly installed in bootstrap.bash because we use it all the time for installing targets?

MatthijsBurgh: I would like to keep bootstrap as clean as possible.

PR #335, opened 2018-12-28
Add SSH support to git clone

ar13pit:


ar13pit:
@jlunenburg May be you could help with this one.

ar13pit:
@MatthijsBurgh I think its done

installer/scripts/tue-install-impl.bash
```diff
@@ -172,12 +172,37 @@ function tue-install-svn
...
 {
+    # Generate SSH url
+    _make-ssh-url-git $1
+
     tue-install-debug "tue-install-git $@"
     if [ ! -d $2 ]; then
         tue-install-debug "git clone --recursive $1 $2"
-        res=$(git clone --recursive $1 $2 2>&1)
+        # If cloning with SSH fails use HTTPS
+        res=$(git clone --recursive $git_repo_ssh_url $2 2> /dev/null || git clone --recursive $1 $2 2>&1)
```
MatthijsBurgh: I would like to see it as an IF which checks a parameter. Would be nice to add tooling for that, so add functions which can set the parameter.

I don't want SSH as the default.

ar13pit: Parameter checking in what sense? 

And what's the harm in having SSH as default? It preferred over HTTPS because it is more secure.

MatthijsBurgh: Few issues:
- What is the influence on speed of trying and failing if you are doing this 100x?
- Once you cloned this way, you can't switch between SSH and HTTPS, which is also limiting.
- In our tue-robocup-* functions we assume the use of HTTPS, see for example [_tue-add-git-remote](https://github.com/tue-robotics/tue-env/blob/master/setup/tue-functions.bash#L732) This would break by merging this.

ar13pit: So I have added an IF condition but thinking about your first issue, do you want me to add a counter that after 5 consecutive failures will issue a warning and just switch to HTTPS?

MatthijsBurgh:
Closing this PR as we have a new implementation in #360 and #361

PR #336, opened 2018-12-29
add gazebo_models

MatthijsBurgh:


installer/targets/gazebo_models/install.bash
```diff
@@ -0,0 +1,10 @@
+tue-install-system-now mercurial
+
+model_path=$HOME/data/gazebo_models
+if [ ! -d $model_path ]
+then
+    hg clone https://bitbucket.org/osrf/gazebo_models $model_path
```
ar13pit: It would be better if you added `tue-install-mercurial` funtion in `tue-install-impl.bash` as well and then use that here.

ar13pit: Well in that case I'll approve this for now.

PR #339, opened 2019-01-04
Support spaces in filenames and create target directory if it does not yet exist

Timple:
Also use the whoami command since the $USER variable is not guaranteed to exist

### DIFF
--- a/installer/scripts/tue-install-impl.bash
+++ b/installer/scripts/tue-install-impl.bash
@@ -284,7 +284,7 @@ function tue-install-cp
 
     # Check if user is allowed to write on target destination
     local root_required=true
-    if namei -l "$2" | grep -q `whoami`
+    if namei -l "$2" | grep -q $(whoami)
     then
         root_required=false
     fi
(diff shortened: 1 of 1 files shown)


installer/scripts/tue-install-impl.bash
```diff
@@ -275,27 +275,29 @@ function tue-install-cp
...
-    if [ ! -f $file ]
+    if [ ! -f "$file" ]
     then
         tue-install-error "Invalid tue-install-cp call: file '$1' does not exist."
     fi
 
     # Check if user is allowed to write on target destination
     local root_required=true
-    if namei -l $2 | grep -q $USER
+    if namei -l "$2" | grep -q `whoami`
```
ar13pit: Change `whoami` to $(whoami) to make it more readable

installer/scripts/tue-install-impl.bash
```diff
@@ -324,7 +326,7 @@ function tue-install-add-text
     local target_file=$2
 
     local root_required=true
-    if namei -l $target_file | grep -q $USER
+    if namei -l $target_file | grep -q `whoami`
```
ar13pit: Change `whoami` to $(whoami) to make it more readable

PR #340, opened 2019-01-04
Add dpkg installation option

Timple:


installer/scripts/tue-install-impl.bash
```diff
@@ -467,6 +468,20 @@ function tue-install-snap
 
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
+function tue-install-dpkg
```
ar13pit: Are we using it anywhere? Not sure if any of the targets have a dpkg command in the install scripts

Timple: Nope, not using it (yet)

installer/scripts/tue-install-impl.bash
```diff
@@ -467,6 +468,20 @@ function tue-install-snap
...
+{
+    if [ -z "$1" ]
+    then
+        tue-install-error "Invalid tue-install-dpkg call: needs package as argument."
+    fi
+    tue-install-debug "Installing dpkg $1"
```
MatthijsBurgh: Maybe check if file exist?

installer/scripts/tue-install-impl.bash
```diff
@@ -467,6 +468,20 @@ function tue-install-snap
...
+    sudo dpkg --install $1
```
MatthijsBurgh: Maybe check if file exist?

installer/scripts/tue-install-impl.bash
MatthijsBurgh: I would also like to see an implementation in the yaml parser. Or is there a reason why this is not done (at nobleo)?

installer/scripts/tue-install-impl.bash
Timple: The error thrown by dpkg is:
```bash
dpkg: error processing archive testszz (--install):
 cannot access archive: No such file or directory
```
Would that suffice?

installer/scripts/tue-install-impl.bash
Timple: That makes sense, I'll can add this if desired/
Here we use it in some internal script.

MatthijsBurgh: Yep that would be nice

installer/scripts/tue-install-impl.bash
MatthijsBurgh: Yep I think so.

installer/scripts/parse-install-yaml.py
```diff
@@ -54,6 +54,8 @@ def main():
                 command = "tue-install-ppa {0}".format(install_item["name"])
             elif install_type == "snap":
                 command = "tue-install-snap {0}".format(install_item["name"])
+            elif install_type == "dpkg":
+                command = "tue-install-dpkg {0}".format(install_item["name"])
```
ar13pit: Could you update the README as well?

MatthijsBurgh:
I will test this in the next few days, not later than saturday.

MatthijsBurgh:
I applied it in the megasync target, please review that commit before merging.

ar13pit:
@MatthijsBurgh Maybe you could make changes to megatools as well. There are some `cp` statements there that should be used with tue-install-cp

PR #343, opened 2019-01-04
Add support for type==empty

Timple:
Some package have dependencies which one might not want to install explicitly. Like `linux-kernel-headers` in [socketcan_interface](https://github.com/ros-industrial/ros_canopen/blob/melodic-devel/socketcan_interface/package.xml)

Therefor I added support for 
```yaml
- type: empty
```


MatthijsBurgh:
This might be a better solution than having an empty target (not having any install.* files)

installer/scripts/tue-install-impl.bash
```diff
@@ -110,7 +110,8 @@ function tue-install-target
         if [ -f $install_file.yaml ]
         then
             tue-install-debug "Parsing $install_file.yaml"
```
MatthijsBurgh: we should throw an error/warning if a target doesn't contain a install.bash/yaml.

Timple: Nah, committing an empty directory is quite hard in git. So someone who does that should know better ;-)
I might even think of a usecase where one would put only a README.md in a target and not a install.bash/yaml to explain what should happen or why this installer cannot cover this item.

MatthijsBurgh: In a case that the there is only a readme. The user should be informed in some way that it should follow the steps in the readme. So this should be in the installer or in the target.

ar13pit:
Before merging, update README as well.

Timple:
Updated readme in dpkg branch. Would have resolved in conflicts anyhow.

MatthijsBurgh:
I will test this in the next few days, not later than saturday.

MatthijsBurgh:
The warning is also thrown when an install.yaml with type empty is available. This is very annoying and should be different IMO.

MatthijsBurgh:
I doesn't throw a warning anymore if you have a correct empty target. So defined in the yaml file.

PR #345, opened 2019-01-12
move keras models to data folder

MatthijsBurgh:


MatthijsBurgh:
See https://github.com/tue-robotics/tue-env-targets/pull/2

PR #347, opened 2019-01-15
Split targets

MatthijsBurgh:
Split targets from implementation

Probably need to built in the CI that it checks out a different branch of the targets if available.


Timple:
Would it be possible to have more target repos?
So one overlaying the other (packages are checked in the target repos in the specified order)

MatthijsBurgh:
That is not in there yet. That would require some more refactoring, but it would be a nice feature.

MatthijsBurgh:
@Timple, I think that would require too much work for now. So let's do that at a later moment.

Timple:
Sure, is fine for later.
Would also be convenient for users to host their own personal targets somewhere.

I'll let @ar13pit review this one since my environment is not correctly setup to test this multi-repo change.

ar13pit:
I'll review this tomorrow

MatthijsBurgh:
To get it working without running the bootstrap
```
TUE_ROS_DISTRO=kinetic
git clone https://github.com/tue-robotics/tue-env-targets.git  ~/ros/$TUE_ROS_DISTRO/.env/targets
echo "export TUE_ENV_TARGETS_DIR=~/ros/$TUE_ROS_DISTRO/.env/targets" >> ~/ros/$TUE_ROS_DISTRO/.env/setup/user_setup.bash
```

ar13pit:
List of functions that break:
1. `tue-robocup-update` in `tue-functions.bash`

Changes needed:
1.  Move cloning of target repo to something like `tue-env` as new environment initialized will have no target repo

MatthijsBurgh:
New conversion script
```
tue-env init-targets ros-$TUE_ROS_DISTRO https://github.com/tue-robotics/tue-env-targets.git
```

MatthijsBurgh:
I don't think we should create a variable for `$TUE_DIR/installer`
I like to keep the directory structure for now. We can change that at a later moment.
If we move the bootstrap to the root of the repo, make it hidden.

installer/tue-install.bash
```diff
@@ -0,0 +1,76 @@
...
+        if [[ ! $REPLY =~ ^[Yy]$ ]]
+        then
+            exit 1
+        fi
+    fi
+fi
+
+if [ ! -d $TUE_ENV_TARGETS_DIR ]
+then
+    git clone https://github.com/tue-robotics/tue-env-targets.git $TUE_ENV_TARGETS_DIR
```
ar13pit: This is bound to cause problems.

ar13pit: May be a good solution is to exit with error instead of cloning

MatthijsBurgh: It is the same as tue-env. But throwing an error would be better

MatthijsBurgh: Done

ar13pit:
So here we proceed with a major upgrade!

MatthijsBurgh:
@ar13pit a PR is not the place to keep notes. So do you put them in a issue. So when we have time for lower prio items we can still find them.

ar13pit:
I completely forgot about moving them. Good that you noticed it else it would have just remained here.

PR #349, opened 2019-01-17
Move functions to bin

MatthijsBurgh:
Move most functions to bin folder, so they don't live in your environment.

merge after #347


ar13pit:
I was trying to do something similar locally but my structure was a bit different. Just one question before we proceed with this:

Why move the functions to bin? Why not to a sub-directory within `setup` and keep `bin` separate only to store compiled binaries rather than scripts.

In what I was trying to do, I wanted to have just one high level script being sourced in `setup.bash` instead of having a loop that is sourcing all the functions and this high level script calling all the functions defined in the other scripts (as you did here). But I got stuck with the autocomplete functions as they became really complicated.

setup.bash
```diff
@@ -66,7 +82,3 @@ if [ -f $TUE_DIR/setup/tue-misc.bash ]
 then
     source $TUE_DIR/setup/tue-misc.bash
```
ar13pit: May be merge this with tue-functions.bash as both just have one function each in them.

MatthijsBurgh:
I will put this on hold for the moment. Shell scripts are run in a child process and therefore don't have access to the environment variables. Another downside of that, is that change of directory don't work, because this only happens in the child process.

One option is create a function which calls the shell script with a dot. Then the script is run in the current process. The downside of this is that some caching takes place. Therefore after running it once, it still 'lives in your environment'. And resourcing is needed after applying changes.

Another option is to end each script with `exec bash`. This exits the current shell and switches to the child process. This does solve the changing of directory issue. I am not sure it will also fix the acces to the environment variables. (I would say it wouldn't). Because it start sourcing the the `.bashrc` after executing the main contents of the script. This takes a lot of time.

I know something similar is used at SR, we could ask @jlunenburg how they solved it/worked around it.

PR #350, opened 2019-01-18
Add target repo suggestion

ar13pit:


README.md
```diff
@@ -12,10 +14,18 @@ tue-make
 source ~/.bashrc
 ```
 
+#### Customization
+A customized targets repository can be setup with this package manager (currently only one git repository is supported). If `tue-env` is already installed, to setup the targets repository run:
+```bash
+tue-env init-targets [ENVIRONMENT] <targets_repo_git_url>
```
Timple: Is anybody actually using these environments?
To me they are a big unknown on how they work or are setup.

They only cause a lot of issue once not correctly initialised

(Not a comment on this PR btw.)

ar13pit: I had the same questions a few months back until I did not dive into the code.
Actually they should be called as workspaces because that is what they really are.

The initialization is just one command and requires only the `tue-env` script. It just creates a directory and assigns a name to it. This name and location is saved inside the root directory of tue-env so that switching between these workspaces is made possible. In a bash session, only one workspace can be active and its contents are sourced by `setup.bash`.

And now that we have separated the targets repo from `tue-env`, each workspace must be associated with a targets repo. However, in the current implementation there is no provision to associate the same targets repo with multiple workspaces.

ar13pit:
@MatthijsBurgh Can you look at this and merge it

PR #354, opened 2019-01-25
Throw error if installed target, doesn't exist

MatthijsBurgh:


Timple:
Should it? It also does not continue to other targets in the list.

If you want to do all of this, simply revert PR. But this makes it hard to find this stuff. 
I'd merge it as is...

installer/tue-install-impl.bash
```diff
@@ -742,7 +742,7 @@ fi
 for target in $targets
 do
     tue-install-debug "Main loop: installing $target"
-    tue-install-target $target
+    tue-install-target $target || tue-install-error "Installed target: '$target' doesn't exist anymore"; return 1
```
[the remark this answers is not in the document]
ar13pit: @MatthijsBurgh Why is there a return statement? Shouldn't there be a continue instead?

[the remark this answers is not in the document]
MatthijsBurgh: No it should stop. So you can take action and run it agian. But it stops too much. So that needs to be fixed.

MatthijsBurgh:
Ready for a new review @Timple @ar13pit

installer/tue-install-impl.bash
```diff
@@ -745,12 +745,19 @@ for target in $targets
...
+        return 1
+    fi
 
     if [[ -d $TUE_INSTALL_TARGETS_DIR/$target && "$tue_cmd" == "install" ]]
     then
         # Mark as installed
         tue-install-debug "[$target] marked as installed after a successful install"
         touch $TUE_INSTALL_INSTALLED_DIR/$target
+    else
+        tue-install-debug "[$target] succesfully updated"
```
ar13pit: I don't understand this

MatthijsBurgh: Oh I missed that

MatthijsBurgh: Done, but shouldn't the checking of the directory not happen before the tue-install-target? Then no checking of the error code is needed.

MatthijsBurgh:
I am going to change it. We just need to check if directory exists

ar13pit:
I want to have a look before you merge this

MatthijsBurgh:
@ar13pit see my changes. I did a general cleanup too.

installer/tue-install-impl.bash
```diff
@@ -744,13 +734,14 @@ fi
 for target in $targets
 do
     tue-install-debug "Main loop: installing $target"
-    tue-install-target $target
+    tue-install-target $target || tue-install-error "Installed target: '$target' doesn't exist (anymore)"
 
-    if [[ -d $TUE_INSTALL_TARGETS_DIR/$target && "$tue_cmd" == "install" ]]
-    then
-        # Mark as installed
+    if [[ "$tue_cmd" == "install" ]]
```
ar13pit: Did you remove the check for target directory because it is already being checked inside the function tue-install-target ?

installer/tue-install-impl.bash
```diff
@@ -36,7 +36,7 @@ Error while installing target '$TUE_INSTALL_CURRENT_TARGET':
 
     $1
 \033[0m"
-    return 1
+    exit 1
```
ar13pit: Do you really want to break out of the installation if one target fails?

installer/tue-install-impl.bash
```diff
@@ -816,14 +810,15 @@ if [ -n "$TUE_INSTALL_SYSTEMS" ]; then
 fi
```
ar13pit: I guess it will be a good idea to add tue-install-error pipe to all the functions that are executed at the end like tue-install-system-now if we are going to break out of the installation

installer/tue-install-impl.bash
MatthijsBurgh: It is best thing to do in my opinion. Because it makes it very clear where the error occurs.

installer/tue-install-impl.bash
MatthijsBurgh: Indeed.

installer/tue-install-impl.bash
MatthijsBurgh: True.

ar13pit: So are you going to do that in this PR?

installer/tue-install-impl.bash
```diff
@@ -841,14 +834,15 @@ if [ -n "$TUE_INSTALL_SYSTEMS" ]; then
...
```
ar13pit: I think tue-install-error pipe must be added to all the functions in this PR itself, as it is an implication of all the changes made here.

MatthijsBurgh: You are free to do so, I don't have the time for it now.

ar13pit:
@MatthijsBurgh Good to merge

PR #358, opened 2019-02-07
Fix version of targets repo in CI [2/2]

### LINKED ISSUE #356: Correct version of targets in CI
At the moment, the targets are not checked out to any branch. Something is implemented, but this is not working correctly.

ar13pit:
fixes #356 

This will surely fix the issue. If @MatthijsBurgh you want to handle error in the git command differently then make the required change.

The change may not be visible instantly as Docker Hub takes some time to have the new image after PRs are merged to master

### DIFF
--- a/installer/tue-install.bash
+++ b/installer/tue-install.bash
@@ -59,22 +59,25 @@ else
 fi
 
 # If the same branch as the PR exists on tue-env-targets, switch to it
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
See my PR #359 for a cleaner solution in my opinion.

PR #360, opened 2019-02-09
SSH support

MatthijsBurgh:
- Including automatic switching of all cloned repos, in both ways.
- Fully supporting all other functions

Thanks to SR @jlunenburg


setup/tue-functions.bash
```diff
@@ -19,6 +19,66 @@ function _list_subdirs
...
+    sed -i \
+        -e '/^#\?\(\s*'"export ${key}"'\s*=\s*\).*/{s//\1'"${value}"'/;:a;n;ba;q}' \
+        -e '$a'"export ${key}"'='"${value}" $3
+}
+
+# ----------------------------------------------------------------------------------------------------
+#                                              SSH
+# ----------------------------------------------------------------------------------------------------
+
+function tue-use-ssh
```
ar13pit: I think this should be moved to tue-env as it is more of an environment property than that of tue-functions.bash

ar13pit:
Before merging this have a look at my PR #361

Timple:
Looks valid. 

It only supports github, but I guess this suits the tue-env just fine.
I am unable to test this right now. So if someone can validate the functionality :)

MatthijsBurgh:
Tested

PR #361, opened 2019-02-10
Move ssh enable to tue-env

ar13pit:
Enhancement to #360.

The implementation of SSH support in @MatthijsBurgh PR make it a property of `tue-env` instead of being just a generic tue function.

I initially moved the SSH implementation to be a part of `tue-env` which Matthjis was against as `tue-env` would keep on getting longer with new functionalities of the environment. However he pointed out all the environments can manually be configured with `tue-env config`. 

So I finally moved the SSH functionality into a new script called `tue-env-config.bash` which can be called by executing `tue-env config [Environment] [ARGS]`


ar13pit:
@Timple your opinion is required on this

Timple:
To be honest I never really got the whole config part. Don't know when to use it. And it depends on environments which only cause me headaches because they error out a lot if I mess up stuff.

I'll leave this one to you two :)

ar13pit:
@MatthijsBurgh What in your opinion are the disadvantages of this PR?

MatthijsBurgh:
- I think we should leave some space for custom config
- So much code for such little functionality
- Needs a lot of work when expanding for another functionality.

ar13pit:
- Custom config has been taken care of in the implementation
 - Agreed its a lot of code for just two functions `use-ssh` and `use-https`, but its taking into account the generalization to add new functionality.
 - Adding new functionality is merely adding a new function and adding its call to the parameter handling part

Anyway I leave it to you to keep it or remove it. In my opinion its a good implementation.

MatthijsBurgh:
@ar13pit now it really dynamic, you can extend the feature by just adding tue-env-XXX functions. Including auto completion

PR #362, opened 2019-02-11
Add option to specify different sources per ROS distro

Timple:
Might be useful when switching to melodic in the future but when not all packages are available yet:
http://repositories.ros.org/status_page/compare_kinetic_melodic.html

### DIFF
--- a/installer/parse-install-yaml.py
+++ b/installer/parse-install-yaml.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python
 
+from os import environ
 import sys
 import yaml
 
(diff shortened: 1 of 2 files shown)


installer/parse-install-yaml.py
```diff
@@ -31,19 +32,32 @@ def main():
 
         try:
             install_type = install_item["type"]
-            if install_type == "ros":
-                source_type = install_item["source"]["type"]
+            if install_type == "empty":
```
MatthijsBurgh: We need to merge the empty target PR first, which has some unresolved issues.

installer/parse-install-yaml.py
```diff
@@ -31,19 +32,32 @@ def main():
...
+                return 0
+            elif install_type == "ros":
+                if "source" in install_item:
+                    source = install_item["source"]
+                else:
+                    release = environ["ROS_DISTRO"]
```
MatthijsBurgh:
```suggestion
                    release = environ["TUE_ROS_DISTRO"]
```
More consistent with rest of the installer

installer/parse-install-yaml.py
Timple: Aaah, that makes sense. I was actually wondering where the Empty statement went anyway.

installer/parse-install-yaml.py
Timple: Suits me, but in which case are they different?

MatthijsBurgh: I don’t think they will be different at any time. But TUE_ROS_DISTRO, will probably be set before ROS_DISTRO, so definitely always available.

PR #363, opened 2019-02-11
Install not often-used tooling on demand

Timple:
Also a bugfix because `software-properties-common` might not be available on every ubuntu install


MatthijsBurgh:
Why not also apply this to git and pip?

Timple:
Thought about it, but it would cause additional delays on initial install. Because they are used often. And this causes a separate install invocation.

So my policy was to only do this for niche products :)

PR #364, opened 2019-02-11
Fix returncode parsing

Timple:


MatthijsBurgh:
Does this fix the issue in the empty target PR?

Timple:
Could be, I forgot what the case was there :)

PR #366, opened 2019-02-12
(tue-env)fix sourcing of env user-setup

MatthijsBurgh:


ar13pit:
I was going to do the same

PR #370, opened 2019-02-14
Remove duplicate ssh functions

ar13pit:
When I merged the enhancement #361 to #360 

These redundancies were added by git. I don't know why this happened.

### DIFF
--- a/setup/tue-functions.bash
+++ b/setup/tue-functions.bash
@@ -19,66 +19,6 @@ function _list_subdirs
     done
 }
 
-function _set_export_option {
-
-    # _set_export_option KEY VALUE FILE
-    # Add the following line: 'export KEY=VALUE' to FILE
-    # Or changes the VALUE to current value if line already in FILE.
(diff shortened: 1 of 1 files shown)


setup/tue-functions.bash
```diff
@@ -19,66 +19,6 @@ function _list_subdirs
...
-
-function tue-use-https
-{
-    local option="TUE_USE_SSH"
-    local value="false"
-    _set_export_option "$option" "$value" $TUE_ENV_DIR/.env/setup/user_setup.bash
-    source $TUE_ENV_DIR/.env/setup/user_setup.bash
-}
-
-function _github_https
```
MatthijsBurgh: All the github functions are needed in the installer. They are also exported so they are available in tue-env config.

ar13pit: Expand the file and below the removed area and you will see they are there. They ended up being defined twice due to the merge

PR #372, opened 2019-02-18
Git Local Housekeeping

### LINKED ISSUE #371: Git housekeeping
With `tue-get update` we prune the remote origin, however we do not touch the local branches.

I would like to add a separate function that cleans up the local branches for a specific package repo or all the package repos based on the option provided by the user.

I won't add this function call to any `tue-get` calls instead will keep it as a separate function all together so that if a user goes into a self-destruct mode, its his problem and not ours.

ar13pit:
Function to remove local stale branches from a specified repo.

closes #371


MatthijsBurgh:
I have a git script which does something similar in MEGA. Please check that one, if it could improve this function.

I also like to have a function which does this for all TUe repo's. Maybe you could use `_tue_repos-do`

ar13pit:
@MatthijsBurgh have a look. I have added things from your script in MEGA and wrapped it for `_tue-repos-do` as well.

ar13pit:
@MatthijsBurgh ping :)

MatthijsBurgh:
busy with report writing....

ar13pit:
@MatthijsBurgh review required

setup/tue-functions.bash
```diff
@@ -19,6 +19,166 @@ function _list_subdirs
...
+function _tue-git-branch-clean
+{
+    # Function to remove stale branches from a git repository (which should
+    # either be the PWD or one of its parent directories). The function removes
+    # stale branches in two layers. First it removes all branches that have been
+    # merged in the remote, then it checks for unmerged branches that have been
+    # deleted from the remote and prompts for confirmation before removal. If
+    # the function is called with "--force-remove" flag, then no confirmation is asked
+
+    local force_remove=
```
MatthijsBurgh: Prefer the declaration with equal sign.

setup/tue-functions.bash
```diff
@@ -19,6 +19,166 @@ function _list_subdirs
...
+        # upon confirmation by the user
+        if [ ! $error_code -eq 0 ]
+        then
+            unmerged_stale_branches="$unmerged_stale_branches $stale_branch"
+        fi
+    done
+
+    # Removal of unmerged stale branches. Not a default operation with the high
+    # level command tue-git-branch-clean
+    if [ -n "$unmerged_stale_branches" ]
```
MatthijsBurgh: I would be annoyed by this. But I don't want to force remove branches.

setup/tue-functions.bash
ar13pit: Did you mean without? Because it is with equal sign here

setup/tue-functions.bash
ar13pit: That's the only way of removing unmerged stale branches. Safe deletion gives an error with such branches. And just for safety I ask for user confirmation. In practice you will see that only few repos will have such branches so the user won't be prompted so many times.

And if the user knows what he's doing, then he can call the function with `--force-remove` to remove the unmerged stale branches

setup/tue-functions.bash
```diff
@@ -19,6 +19,166 @@ function _list_subdirs
...
+    # stale branches in two layers. First it removes all branches that have been
+    # merged in the remote, then it checks for unmerged branches that have been
+    # deleted from the remote and prompts for confirmation before removal. If
+    # the function is called with "--force-remove" flag, then no confirmation is asked
+
+    local force_remove=
+    local error_code=
+    local stale_branches=
+    local repo=
+    repo="$PWD"
```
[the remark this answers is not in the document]
ar13pit: I think it is if you use `_tue-repos-do` because that command first does `cd` into the directory and then evaluate the string command.

Taking that into account, I have written this function in a way that it is executed only inside a git repo. If you look at the statement of `git fetch -p`, then you will see that the function exits if `git fetch -p` fails (one of the reason also includes that the directory is not a git repo)

setup/tue-functions.bash
MatthijsBurgh: without*

setup/tue-functions.bash
ar13pit: I think I have a nice solution that wouldn't be annoying. If there are unmerged stale branches, I'll just print out the names of the branches and instead of asking for user input, I'll echo to remove these stale branches, call the command with `--force-remove`.

What do you think about this?

jlunenburg:
@ar13pit @MatthijsBurgh urgent?

MatthijsBurgh:
Yes, this would help to fix a lot of branches stuff.

setup/tue-functions.bash
MatthijsBurgh: Did you implement this?

setup/tue-functions.bash
```diff
@@ -47,16 +47,21 @@ function _tue-git-branch-clean
...
     local repo
-    repo="$PWD"
+    repo=$(basename "$repo_path")
 
     if [ -n "$1" ]
     then
         if [ "$1" == "--force-remove" ]
+        then
+            force_remove=true
+        elif [ -z "$2" ] && [ "$1" == "true" ]
```
ar13pit: What's with this? Why do you want to have a parameter "`true`"? Having just an option like `--force-remove` is more than enough

setup/tue-functions.bash
```diff
@@ -104,15 +105,15 @@ function _tue-git-branch-clean
...
+            unmerged_stale_branches="${unmerged_stale_branches:+${unmerged_stale_branches} } $stale_branch"
         fi
     done
 
     # Removal of unmerged stale branches. Not a default operation with the high
     # level command tue-git-branch-clean
     if [ -n "$unmerged_stale_branches" ]
     then
-        unmerged_stale_branches=$(echo "$unmerged_stale_branches" | tr " " "\n")
+        unmerged_stale_branches=$(echo "$unmerged_stale_branches" | sed -e 's/^[[:space:]]*//' | tr " " "\n")
```
ar13pit: Is the `sed` command really required because when I last tested this, it worked

setup/tue-functions.bash
ar13pit: Yup. I think you know it by now :laughing:

setup/tue-functions.bash
MatthijsBurgh: Because you pass on true from `tue-git-branch-clean` to `_tue-git-branch-clean`. So either this way, or just pass on all arguments to the child function.

ar13pit: I think passing on all the arguments to the child function is cleaner. This looks ugly. I definitely made a mistake by parsing the argument in the parent function.

ar13pit: However a simple check in the parent function is required else there will be tons of printed statements on the screen.

setup/tue-functions.bash
MatthijsBurgh: yep, otherwise there is an empty lime between unmerged stale branches and the actual branches.

MatthijsBurgh:
I am still doubting `tue-git-branch-clean` is the best name. I would suggest `tue-git-clean-local-branches`

PR #374, opened 2019-02-22
Let log be identical to output

Timple:
Current log is not very useful and could be much more extensive.
Made it identical to stdout so you don't have to understand multiple ways of logging things as user.


MatthijsBurgh:
Looks good to me, but how does it work with colors? Otherwise you get a very messy log file, with all these weird characters.

Timple:
Works fine if you just use `cat`.

PR #368, opened 2019-02-13
Docker and CI cleanup

ar13pit:
- Reduce the layers in the docker image
- CI scripts cleanup
- Execute CI scripts only in CI environment

PR #376, opened 2019-03-05
Fix local builds of Docker image

ar13pit:
In my previous PR #368 I did some cleanup of the Dockerfile to reduce the number of layers in the docker image. 

However, the local builds were still an issue as the entire repository directory was being copied. In this PR I have made the installation of `tue-env` strictly inside the docker image by cloning the repository inside it rather than copying into it. This also ensures that `tue-env` installation procedure is identical in:
1. Standard Installation
2. In CI environment ~~(Push request)~~
3. In local Docker image builds

CI branches have been taken care of by passing it as a build argument to the Dockerfile. The default value has been set to master (this will be the most common local build use case).

~~In case of PR, the Dockerfile is dynamically changed to add a local copy of the repo to the image instead of cloning and the existing procedure to build the image for PR is followed.~~

**Update:** With the current implementation, `tue-env` repository will not be in a detached state in the docker image. It will be checked out the branch corresponding to the image tag


Dockerfile
```diff
@@ -30,16 +39,15 @@ RUN apt-get update -qq && \
...
 
 # Setup tue-env and install target ros
     # Remove interactive check from bashrc, otherwise bashrc refuses to execute
 RUN sed -e s/return//g -i ~/.bashrc && \
-    # Change the ownership of tue dir to USER
-    sudo chown -R "$USER:$USER" ~/.tue/ && \
-    # Run the standard installation script
-    ~/.tue/installer/bootstrap.bash && \
+    # Run the standard installation script if not in a PR
+    source <(wget -q -O - https://raw.githubusercontent.com/tue-robotics/tue-env/"$CI_BRANCH"/ci/docker-build-select.bash) && \
```
ar13pit: This is more of a bash thing than Docker.
@MatthijsBurgh @Timple The only thing I am a bit doubtful about is the branch I select for build selection.

installer/bootstrap.bash
```diff
@@ -40,9 +40,26 @@ then
...
-if [[ -z "$CI" ]]
+# If in CI with Docker, then clone tue-env with CI_BRANCH
+if [ -n "$CI" -a -n "$DOCKER" ]
 then
+    if [ "$CI_PULL_REQUEST" == "false" ]
+    then
+        if [ -n "$CI_BRANCH" ]
+        then
+            echo -en "[tue-env](bootstrap) Docker CI environment found"
+            echo -en "[tue-env](bootstrap) Cloning tue-env repository with branch: $CI_BRANCH"
```
ar13pit: In the log somehow these lines are not being printed in separate lines.

ar13pit:
In the case of PR, the build is done from the local repository on Travis:
https://travis-ci.org/tue-robotics/tue-env/builds/502554099#L567

In the case of no PR (i.e., Push) the build is done by cloning inside the image:
https://travis-ci.org/tue-robotics/tue-env/builds/502554081#L509

ar13pit:
All of my changes take into account the existing implementation. The ideology behind the existing implementation has been preserved, i.e. push a new docker image with the PR_BRANCH tag only when in the push mode and in PR mode test the build in the master branch.

With the changes, Docker images built locally on our laptops are now independent of our local tue-env repo and can be checked out to a specific branch during the build by using the construct:
`docker build --build-arg BRANCH=<desired_branch> -t <IMAGE_NAME:TAG> .`

ar13pit:
@MatthijsBurgh This ready :)

ar13pit:
I think we can do a git pull in the `tue-install.bash` for `tue-get` instead of git fetch as the docker container will never be in a detached state now.

But that's for a different PR.

ar13pit:
@MatthijsBurgh ping :)

PR #377, opened 2019-03-07
Set a default data directory

ar13pit:
Directory to save models for image and speech recognition

Before merging correct the paths in targets (tue-robotics/tue-env-targets#29)

### DIFF
--- a/setup.bash
+++ b/setup.bash
@@ -9,7 +9,8 @@ source $TUE_DIR/setup/tue-env.bash
 function _tue-check-env-vars
 {
     [ -n "$TUE_DIR" ] && [ -n "$TUE_ENV" ] && [ -n "$TUE_ENV_DIR" ] \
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
Check https://github.com/tue-robotics/tue-env/blob/master/setup/tue-data.bash. Data has always been stored in `~/ros/data(/private)`, so keep it there.

ar13pit:
I did see that but the keras models are being downloaded to a different location
https://github.com/tue-robotics/tue-env-targets/blob/master/ros-image_recognition_keras/install.bash#L3

So thought of generalizing the location. But the question then is why is the keras model not being placed there?

MatthijsBurgh:
See updated comment, now I will be offline again.

ar13pit:
Closing this PR as this issue needs more discussion and maybe a new design for data management of an env workspace.

PR #380, opened 2019-03-10
Switch back to tuerobotics organization os dockerhub

MatthijsBurgh:


ar13pit:
Did you find the ID and password of that account?

MatthijsBurgh:
tuerobotics is an organization. We used tueamigo account to access it, we lost the credentials of that account. But @Timple was also member of the organization. So he added the tueroboticsamigo user.

MatthijsBurgh:
Because it is better to have an organization and nice to have the name matching the github organization.

PR #384, opened 2019-03-25
Fix CI install package

ar13pit:
Merge after CI for tue-robocup shows target branch pull


ar13pit:
For reference https://travis-ci.org/tue-robotics/tue_robocup/builds/511192541#L524

PR #385, opened 2019-03-28
Use rosdep

MatthijsBurgh:
Use rosdep to resolve targets, which are not in the list.

(So you can overrule rosdep by adding a target.)

Please only approve, **don't merge**.

Review together with https://github.com/tue-robotics/tue-env-targets/pull/43

### DIFF
--- a/setup/tue-functions.bash
+++ b/setup/tue-functions.bash
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-_TUE_CATKIN_DEV_DIR=$TUE_ENV_DIR/dev
-_TUE_CATKIN_SYSTEM_DIR=$TUE_ENV_DIR/system
-
 # ----------------------------------------------------------------------------------------------------
 #                                        HELPER FUNCTIONS
 # ----------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


LoyVanBeek:
Please test the everloving sh*t out of this before merging, will break a lot of installs if not I'm afraid.

MatthijsBurgh:
That is why I added "please don't merge" line. I will only merge this after a sh*tload of testing. Already have some issues, which needs to be addressed.

Timple:
Why this implementation of checking every dependency? Why not just do a massive rosdep-install this-dir at the end?

MatthijsBurgh:
Then you register which dependencies are met during the main install procedure and then check which are met during the rosdep-install-this-dir. Which gets pretty messy I would say.

Timple:
Why do you need to check what rosdep resolves? Just install everything you know (including override targets) and end with a massive rosdep install all-that-is-left. 

And if that fails there is probably a nice error message in place anyway.

MatthijsBurgh:
> Why do you need to check what rosdep resolves? Just install everything you know (including override targets) and end with a massive `rosdep install all-that-is-left`.
> 
> And if that fails there is probably a nice error message in place anyway.

I will take that in consideration.

MatthijsBurgh:
The current issue which remains a high priority. How do you guarantee that rosdep is available. As it is installed by the ros target at the moment. But because `tue-install-rosdep` is now called from `tue-install-target`, there is no guarantee in having `rosdep` aka having ros target installed. If you move the call to tue-install-ros, then ros target is installed, so `rosdep` is available. But this removes the possibility to install any ros dependency directly from `tue-get`. So `tue-get install std_msgs` will not be possible anymore. (By removing the targets, auto completion will also be removed for these targets)

Timple:
Just make rosdep a dependency of ros?

And ros is always installed before any ros target.

MatthijsBurgh:
> Just make rosdep a dependency of ros?
> 
> And ros is always installed before any ros target.

So move the `tue-install-rosdep` to `tue-install-ros`? Because now it is called in `tue-install-target`.

ar13pit:
> > Just make rosdep a dependency of ros?
> > And ros is always installed before any ros target.
> 
> So move the `tue-install-rosdep` to `tue-install-ros`? Because now it is called in `tue-install-target`.

I think what @Timple means is to put rosdep as a dependency in the ros target if I understood him correctly

setup/tue-functions.bash
```diff
@@ -56,36 +53,23 @@ export -f _github_https_or_ssh # otherwise not available in sourced files
 
 function tue-make
 {
-    if [ -n "$TUE_ROS_DISTRO" ] && [ -d $_TUE_CATKIN_SYSTEM_DIR ]
+    if [ ! -d $TUE_SYSTEM_DIR ]
```
ar13pit: In which script is `TUE_SYSTEM_DIR` being defined?

MatthijsBurgh: See below; ros target

PR #387, opened 2019-04-03
Extend 'add text'

MatthijsBurgh:
- Add possibility for relative and absolute source files
- Check if changing the lines, changes the target file. This reduces the need for sudo.
- Add debugging statements

Also fix all shellcheck errors

Maybe best to check each commit separate

### DIFF
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -378,12 +378,11 @@ function tue-install-add-text
     tue-install-debug "tue-install-add-text $*"
 
     local source_file=$1
-    if [[ "$source_file" != "/"* ]] || [[ "$source_file" != "~/"* ]]
+    if [[ "$source_file" == "/"* ]] || [[ "$source_file" == "~/"* ]]
     then
-        tue-install-debug "tue-install-add-text: relative source file"
-        source_file="$TUE_INSTALL_CURRENT_TARGET_DIR/$source_file"
+        tue-install-error "tue-install-add-text: Only relative source files to the target directory are allowed"
     else
-        tue-install-debug "tue-install-add-text: absolute source file"
+        source_file="$TUE_INSTALL_CURRENT_TARGET_DIR/$source_file"
     fi
     local target_file=$2
     if [[ "$target_file" != "/"* ]] && [[ "$source_file" != "~/"* ]]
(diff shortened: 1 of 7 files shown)


installer/tue-install-impl.bash
```diff
@@ -375,47 +375,74 @@ function tue-install-add-text
...
+        if ! cmp --quiet "$tmp_target_file" "$target_file"
         then
-            sed -e "/^$end_tag/r $tmp_source_file" -e "/^$begin_tag/,/^$end_tag/d" $target_file | sudo tee $target_file.tmp
-            sudo mv $target_file.tmp $target_file
+            tue-install-debug "tue-install-add-text: Lines are changed, so copying"
+            if $root_required
+            then
+                sudo mv $tmp_target_file $target_file
+            else
+                mv $target_file.tmp $target_file
```
ar13pit: Error in temp filename

installer/tue-install-impl.bash
```diff
@@ -375,47 +375,74 @@ function tue-install-add-text
...
     local target_file=$2
+    if [[ "$target_file" != "/"* ]] && [[ "$source_file" != "~/"* ]]
+    then
+        tue-install-error "tue-install-add-text: target file needs to be absolute or relative to the home directory"
+    fi
 
     local root_required=true
     if namei -l $target_file | grep -q $(whoami)
     then
+        tue-install-debug "tue-install-add-text: NO root required"
```
ar13pit: Check whether root is required to read source file as the source file is not restricted to just the targets repo

installer/tue-install-impl.bash
```diff
@@ -375,47 +375,74 @@ function tue-install-add-text
...
     if [ ! -f $target_file ]
     then
-        tue-install-error "tue-install-add-text: No such file: $target_file"
+        tue-install-error "tue-install-add-text: No such target file: $target_file"
     fi
 
     local begin_tag=$(head -n 1 $source_file)
     local end_tag=$(awk '/./{line=$0} END{print line}' $source_file)
     local text=$(sed -e :a -e '/^\n*$/{$d;N;};/\n&/ba' $source_file)
+    tue-install-debug "tue-install-add-text: Lines to be added: \n$text"
```
ar13pit: If root is required to read source, then `sudo` needs to be added in the above commands

installer/tue-install-impl.bash
MatthijsBurgh: fixed

PR #388, opened 2019-04-04
Add pip3 support

ar13pit:


ar13pit:
@MatthijsBurgh Need approval again. There definitely will be conflict between the executables like ipython but there won't be any conflicts with `import` in the scripts

MatthijsBurgh:
@ar13pit How do prevent/control such conflicts?

ar13pit:
I only faced the conflict problem for ipython. The solution is correcting the symlinks in /usr/bin.

`pip` commands have a tendency to overwrite symlinks if an executable is being installed through them.

So the target that has such a problem, we will have to add this symlink correction in its `setup` file.

ar13pit:
@MatthijsBurgh I found the solution to the conflicts. We need to use explicit versions for example if we install ipython from pip3 then it symlinks ipython3 to ipython. If now we install ipython from pip, it symlinks ipython2 to ipython. In either case, ipython2 and ipython3 remain unaffected.

Also, the symlink change happens only if pip install is called with upgrade option which we do not do.

MatthijsBurgh:
> @MatthijsBurgh I found the solution to the conflicts. We need to use explicit versions for example if we install ipython from pip3 then it symlinks ipython3 to ipython. If now we install ipython from pip, it symlinks ipython2 to ipython. In either case, ipython2 and ipython3 remain unaffected.
> 
> Also, the symlink change happens only if pip install is called with upgrade option which we do not do.
Does this apply to all executables of pip packages?

ar13pit:
This is ready to be merged.

MatthijsBurgh:
I will test this. As I don't want to have a similar situation as we had on HERO2.

ar13pit:
> I will test this. As I don't want to have a similar situation as we had on HERO2.

What happened in HERO2?

MatthijsBurgh:
> > I will test this. As I don't want to have a similar situation as we had on HERO2.
> 
> What happened in HERO2?

I don't know exactly, but it was a mess with python2.7/3 and their pips.

PR #389, opened 2019-04-10
Explicitly use pip2 instead of pip

ar13pit:


ar13pit:
Do test this by removing all installations of pip. It works on my computer but I would like to be really sure before this is merged

PR #392, opened 2019-04-15
Remove auto check for cuda in installer.

MatthijsBurgh:
Let it be decided by the user self. So the user can set TUE_CUDA to use cuda in targets or not.

@LarsJanssenTUe This will prevent the issue tensorflow issue to happen again by running `tue-get update`


ar13pit:
This anyway should be in `user_setup.bash` and not here

PR #394, opened 2019-04-16
exit on error if CI in tue-install.bash

### LINKED ISSUE #393: build-docker-image doesn't stop on error
Doesn't stop when failing to download apt packages. See https://travis-ci.com/tue-robotics/tue-env/jobs/193292094#L543

MatthijsBurgh:
It works: https://travis-ci.com/tue-robotics/tue-env/jobs/193368855

fixes #393

### DIFF
--- a/installer/tue-install.bash
+++ b/installer/tue-install.bash
@@ -108,6 +108,5 @@ then
     fi
 fi
 
-[ -n "$CI" ] && echo -e "[tue-get] Going to install packages '$@'"
 # Run installer
 source $TUE_DIR/installer/tue-install-impl.bash "$@"
--- a/installer/tue-install.bash
+++ b/installer/tue-install.bash
@@ -1,6 +1,8 @@
 #!/bin/bash
 _tue-check-env-vars || return 1
 
+[[ -n "$CI" ]] && set -o errexit
+
 # Update installer
 if [ ! -d $TUE_DIR ]
 then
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
All branch stuff breaks the CI. Because it produces a non-zero error code if that branch doesn't exist.

ar13pit:
I don't understand why is the CI breaks after `sudo apt-get install --assume-yes python-yaml python-pip`

MatthijsBurgh:
> I don't understand why is the CI breaks after `sudo apt-get install --assume-yes python-yaml python-pip`

I think as it as one package

installer/tue-install.bash
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 _tue-check-env-vars || return 1
 
+[[ -n "$CI" ]] && set -o errexit
```
ar13pit: This is wrong. The script breaks out because the branch of `tue-env` does not exist in `tue-env-targets`. The branch of `tue-env-targets` must be set to master in this case and `tue-install-impl` must be executed. Any errors in that script must break the CI.

MatthijsBurgh:
> All branch stuff breaks the CI. Because it produces a non-zero error code if that branch doesn't exist.

@ar13pit I already found out myself.

jlunenburg:
@MatthijsBurgh Urgent at this moment? Else: please move to backlog

MatthijsBurgh:
> @MatthijsBurgh Urgent at this moment? Else: please move to backlog

Nope

PR #395, opened 2019-04-19
Github get/create releases

### LINKED ISSUE #379: Add get github releases
Generalize the get releases script from `ros-dashboard` target and add it `tue-functions.bash`.

Add auto release script also.

ar13pit:
Closes #379 

After merge make appropriate changes to `ros-speech_recognition` and `ros-dashboard` targets from tue-robotics/tue-env-targets#30

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,7 @@
 root = true
 
 [*]
+max_line_length = 120
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ generally used methods of installing packages:
 | `tue-install-svn`               | To install a svn repository                                                                 |
 | `tue-install-system`            | To add `deb` package to a list of packages to be installed at the end with `apt-get`        |
 | `tue-install-system-now`        | To install `deb` packages with `apt-get` right away, but ignores it if already installed    |
+| `tue-install-get-releases`      | To get a released asset from a github repository and place it in the requested directory    |
 
 The input arguments for each of the above mentioned commands can be found by
 simply executing the command in a bash session (provided tue-env is correctly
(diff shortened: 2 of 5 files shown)


ar13pit:
@MatthijsBurgh I have tested this locally for downloading releases from github and it works nicely. It has support for regex for the file name. Test it out for `dashboard` and `hero-display`

1. `hero-display`. Here the latest release is used by default.
    ```bash
    tue-install-get-releases "tue-robotics/hero-display" "^hero.*Image" "."
    ```
1. `dashboard`. Here a release from a specific tag is downloaded
    ```bash
    tue-install-get-releases "tue-robotics/dashboard" "^dashboard(-,+)?-.*.tar.gz" "." "2017-before-rgo"
    ```

The bash function at the moment is just a template function. Checks need to be implemented in it.

MatthijsBurgh:
So how do we get the vX and vX-alpha release?

ar13pit:
The syntax of the command is:
```bash
tue-install-get-releases <REPO_SHORT_URL> <FILENAME> <DOWNLOAD_DIR> [TAG]
```
`FILENAME` supports regular expressions.
`TAG` is optional. If not specified, then the file is downloaded from the latest release.

As vX and vX-alpha are tags, these should be used as the arg 4 of the command.

MatthijsBurgh:
So I could just use this regex, `/^v[0-9]+.[0-9]+.[0-9]+$/` to get a vX.X.X version with x being a number?

ar13pit:
That won't work. I have not added support for regex in tags. At the moment tags need to be exact. So if I understand correctly, the feature that you want is to have regex for filename and tags and then download the most recent file that matches both, if it is not present at the specified output location?

MatthijsBurgh:
Yep, As it would be nice to use this to replace for example `hero-dashboard`, https://github.com/tue-robotics/tue-env-targets/tree/master/hero-dashboard

ar13pit:
Yeah I referred to that itself when I asked the question. This can be extended to meet that requirement. I'll do that tomorrow.

PR #396, opened 2019-04-23
[tue-env] Fix user_setup.bash creation in new env

ar13pit:


MatthijsBurgh:
What is wrong with the current solution?

ar13pit:
If you initialize a new environment and then do `tue-env config` an error is produced because `user_setup.bash` does not exist and the `edit` does not create a blank file if no file exists
```
[~]
arpit@A3-PC $ tue-env config 
Error: no write permission for file "/home/arpit/test/.env/setup/user_setup.bash"
bash: /home/arpit/test/.env/setup/user_setup.bash: No such file or directory
```

MatthijsBurgh:
Would be better to do this when calling `tue-env config` instead of `tue-env init`

ar13pit:
> Would be better to do this when calling `tue-env config` instead of `tue-env init`

So you want to keep explicit creation of the directory and `user_setup.bash` in bootstrap?
It is anyway required for setting up either `ROS_DISTRO` or `USE_SSH` so why not create it anyway with the initialization of a new environment?

jlunenburg:
@MatthijsBurgh @ar13pit is this urgent?

MatthijsBurgh:
> @MatthijsBurgh @ar13pit is this urgent?

Nope

PR #404, opened 2019-05-21
RGO2019

ar13pit:
Added CUDA dockerfile 
Fixes tue-robotics/tue_robocup#806


ar13pit:
@MatthijsBurgh RTM.

MatthijsBurgh:
I am not sure the cuda docker files should be here. Please explain why it should be here?

ar13pit:
With CUDA dockerfile we can run CUDA dependent packages like openpose, openface, keras etc in docker containers. This will not affect the CUDA installation required system wide as the version required for running each of the packages will be restricted to only Docker.

MatthijsBurgh:
But why here?

ar13pit:
For clarity the CUDA docker files prepare a Docker image with CUDA, CuDNN and tue-env installed that's why it is here.

MatthijsBurgh:
Am I correct, there is a much overlap in the dockerfiles. Could this duplication be removed? By having both containers be build from multiple docker files?

ar13pit:
The problem is the base images are different and everything else is the same. This is a way of building multiple images from the same Dockerfile but its complex and not well documented. 

As our build time isn't that much and as non CUDA docker image has build priority, I think for now we shouldn't put time into figuring out the complex thing.

ar13pit:
Enhancement issue #405 created.

PR #407, opened 2019-06-17
Update git-clean-local echo statements

ar13pit:
Added colors, spacing and made sure the name of a removed branch is displayed only after a branch has been removed.


MatthijsBurgh:
Will test this tomorrow

MatthijsBurgh:
When It only found unmerged stale branches, which need the `--force-remove`, it still print the `Removing stale branches:` without any  branches. So either print all branches to be removed there or don't print this if there are only branches which require the `--force-remove`

ar13pit:
@MatthijsBurgh this is done

ar13pit:
> Line 111 also in the IF?

That's fine because the IF block should execute only once while the Line 111 needs to execute multiple times.

PR #408, opened 2019-06-17
Check and display all removed targets first

### LINKED ISSUE #403: Check and display all removed targets first
As the removed targets break the installation/update, this is annoying when multiple removed targets are present.

A solution to this is to first check if installed targets no longer exist and display all such targets and break.

ar13pit:
Lets say we have 100 targets are installed and 5 of them are not in targets repo anymore.

Currently there is no order in which their existence is checked. 

It may so happen that the first missing target is found after updating 20 targets and the installation will break. After fixing this missing target, the installer will again go through all the 20 targets and so on, which is a waste of time. 

This PR fixes this issue.
Fixes #403

### DIFF
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -755,7 +755,7 @@ function _missing_targets_check
     if [ -n "$missing_targets" ]
     then
         missing_targets=$(echo "$missing_targets" | tr " " "\n" | sort)
-        echo -e "\e[31m Error! The following installed targets do not exist (anymore):\n$missing_targets \e[0m"
+        echo -e "\e[31mThe following installed targets don't exist (anymore):\n$missing_targets \e[0m"
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
Will test this tomorrow

installer/tue-install-impl.bash
```diff
@@ -735,6 +735,33 @@ function generate_setup_file
...
+        if [ ! -d "$TUE_INSTALL_TARGETS_DIR"/"$target" ]
+        then
+            missing_targets="$target${missing_targets:+ ${missing_targets}}"
+        fi
+    done
+
+    if [ -n "$missing_targets" ]
+    then
+        missing_targets=$(echo "$missing_targets" | tr " " "\n" | sort)
+        echo -e "\e[31m Error! The following installed targets do not exist (anymore):\n$missing_targets \e[0m"
```
MatthijsBurgh:
```suggestion
        echo -e "\e[31mThe following installed targets do not exist (anymore):\n$missing_targets \e[0m"
```

ar13pit:
> https://github.com/tue-robotics/tue-env/pull/408/files#diff-27df6d20d171684702149fd4165c21e7R831 Is this error catching still needed/correct?

I would keep it as a safety check, although after the new missing targets check there should be no scenario where it must be executed. However if it did, then someone really screwed up his env.

PR #411, opened 2019-08-11
Change CI for multi-package repo

MatthijsBurgh:
Travis runs easily into a time limit when running on a multi-package repo. Therefore the result is actually useless as not all package can be tested.
Another flaw is that success is only determined on the success of the last package. See https://travis-ci.com/tue-robotics/navigation/builds/122691895, `amcl` fails on `map_server` (and many more errors), but because `voxel_grid` succeedes, the whole CI succeedes. This is completely incorrect. 

My proposed setup, runs all the packages in parallel by using the matrix functionality of travis. This prevents running into a timeout, because of running packages serial. (Single packages could still run in a timeout). Also the problem of a incorrect success status is fixed. Packages which haven't changed will not be installed and build. Those packages will succeed in a very short time. (Showing these as ignored isn't possible.)

In the new setup, the travis config needs to be adjusted for each repo. (Travis doesn't allow the matrix to configured on runtime.) All packages need to be added to the list of environment variables. Though this needs to be only done once.

The result of the new CI can be seen here, tested on navigation: https://travis-ci.com/tue-robotics/navigation/builds/122705133

Skipping a package, costmap_2d, is shown here: https://travis-ci.com/tue-robotics/navigation/jobs/224172024
(It was manually skipped by adding it as an argument to `packages.sh`

If have also tested the new scripts on a single package repo, ED. It still works: https://travis-ci.com/tue-robotics/ed/builds/122706565

### DIFF
--- a/ci/.travis.yml.multiple_pkgs.example
+++ b/ci/.travis.yml.multiple_pkgs.example
@@ -12,9 +12,13 @@ before_install:
   # Determine packages modified in this commit
   - source packages.sh # add all folders which need to be excluded as an argument
 
+env:
+  - PACKAGE=package1
+  - PACKAGE=package2
+  - PACKAGE=package3
+
+install:
+  - bash install-package.sh --package=$PACKAGE --branch=${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH} --commit=$TRAVIS_COMMIT --pullrequest=$TRAVIS_PULL_REQUEST
+
 script:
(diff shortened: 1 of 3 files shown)


ar13pit:
How will packages not changed be not built? Do we need to specify the packages to be skipped in the config? Or does the `source packages.sh` automatically do it?

MatthijsBurgh:
`packages.sh` determines which packages need to be build. These will be put in `PACKAGES`. So if the package is not in `PACKAGES` both the install and build script will just exit with 0.

PR #415, opened 2019-09-05
Fix release build

MatthijsBurgh:
- Also build release in CI
- Remove `--cmake-args` as it not needed and broke the usage of additional arguments
- Add build type to all build functions

Merge together with https://github.com/tue-robotics/tue-env-targets/pull/83

### DIFF
--- a/ci/build-package.sh
+++ b/ci/build-package.sh
@@ -42,12 +42,12 @@ fi
 # Use docker environment variables in all exec commands instead of script variables
 # Compile the package
 echo -e "\e[35m\e[1m Compile the package (catkin build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCATKIN_ENABLE_TESTING=OFF) \e[0m"
-docker exec tue-env bash -c 'source ~/.bashrc; cd ~/ros/"$ROS_DISTRO"/system/src/"$PACKAGE" && catkin build --this --no-status -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCATKIN_ENABLE_TESTING=OFF'
+docker exec tue-env bash -c 'source ~/.bashrc; cd "$TUE_SYSTEM_DIR"/src/"$PACKAGE" && catkin build --this --no-status -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCATKIN_ENABLE_TESTING=OFF'
 
 # Run unit tests
 echo -e "\e[35m\e[1m Run tests on this package (catkin run_tests --this --no-deps -DCATKIN_ENABLE_TESTING=ON) \e[0m"
-docker exec tue-env bash -c 'source ~/.bashrc; cd ~/ros/"$ROS_DISTRO"/system/src/"$PACKAGE" && catkin run_tests --this --no-status --no-deps -DCATKIN_ENABLE_TESTING=ON'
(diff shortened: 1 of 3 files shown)


Timple:
This will still brake if people try to do `catkin build` or `catkin build --this` manually.

Why not set the option in the build configuration once?
`catkin config change this`

You can run it once. People can modify after if they like. All builds afterwards will use the same options, so preventing a lot of re-compiling because of different arguments

ci/build-package.sh
```diff
@@ -41,8 +41,8 @@ fi
 
 # Use docker environment variables in all exec commands instead of script variables
 # Compile the package
-echo -e "\e[35m\e[1m Compile the package (catkin build -DCATKIN_ENABLE_TESTING=OFF) \e[0m"
-docker exec tue-env bash -c 'source ~/.bashrc; cd ~/ros/"$ROS_DISTRO"/system/src/"$PACKAGE" && catkin build --this --no-status -DCATKIN_ENABLE_TESTING=OFF'
+echo -e "\e[35m\e[1m Compile the package (catkin build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCATKIN_ENABLE_TESTING=OFF) \e[0m"
```
Timple: Dit hoeft dus niet meer

MatthijsBurgh: Idd

MatthijsBurgh: As discussed, we keep this for clearity

setup/tue-functions.bash
```diff
@@ -223,89 +226,60 @@ export -f _github_https_or_ssh # otherwise not available in sourced files
...
 function tue-make
 {
-    if [ -n "$TUE_ROS_DISTRO" ] && [ -d "$_TUE_CATKIN_SYSTEM_DIR" ]
+    if [ -n "$TUE_ROS_DISTRO" ] && [ -d "$TUE_SYSTEM_DIR" ]
     then
-        case $(cat "$_TUE_CATKIN_SYSTEM_DIR"/devel/.built_by) in
+        case $(cat "$TUE_SYSTEM_DIR"/devel/.built_by) in
         'catkin_make')
-            catkin_make --directory "$_TUE_CATKIN_SYSTEM_DIR" -DCMAKE_BUILD_TYPE=RelWithDebInfo "$@"
+            echo -e "\e[33mcatkin_make is not supported anymore, use catkin tools\e[0m"
```
ar13pit: Still want to have this? I think we can just have a `*` arg for which the `echo` will say the tool is not supported.
Also good to have a `return 1`

setup/tue-functions.bash
```diff
@@ -223,89 +226,60 @@ export -f _github_https_or_ssh # otherwise not available in sourced files
...
-    '')
-        catkin init --workspace "$_TUE_CATKIN_DEV_DIR" "$@"
-        catkin build --workspace "$_TUE_CATKIN_DEV_DIR" "$@"
-        ;;
-    esac
+	if [ -n "$TUE_ROS_DISTRO" ] && [ -d "$TUE_DEV_DIR" ]
+    then
+		case $(cat "$TUE_DEV_DIR"/devel/.built_by) in
+		'catkin_make')
+			echo -e "\e[33mcatkin_make is not supported anymore, use catkin tools\e[0m"
```
ar13pit: Same here.

setup/tue-functions.bash
MatthijsBurgh: I don't know what will happen when the file doesn't exist, then it should continue to creating the workspace.

PR #418, opened 2019-09-10
[tue-apt-select-mirror] new function

MatthijsBurgh:


MatthijsBurgh:
This is just a first version in my opinion. Just like to get your opinion @ar13pit

ar13pit:
@MatthijsBurgh seems to be in the right direction.

setup/tue-functions.bash
```diff
@@ -22,6 +22,37 @@ function _list_subdirs
...
+    # All Arguments to this functions are passed on to apt-select, so check the
+    # apt-select documentation for all options.
+	hash pip2 2> /dev/null|| sudo apt-get install python-pip
+	hash apt-select 2> /dev/null|| pip2 install --user apt-select
+
+	local mem_pwd=$PWD
+    # shellcheck disable=SC2164
+	cd /tmp
+	local err_code
+	apt-select "$@" 1> /dev/null
```
ar13pit: :laughing: :heart:

ar13pit:
New PR needed for adding this function to Docker file or `tue-install.bash` ? Or want to do it in this itself?

setup/tue-functions.bash
```diff
@@ -22,6 +22,37 @@ function _list_subdirs
...
+	err_code=$?
+	if [ $err_code == 4 ]
+	then
+		echo -e "Fastest apt mirror is the current one"
+	else
```
ar13pit: This has chances of failing. If the command has a different error code, `sources.list` will not be created and the `else` block will start spitting out errors.

PR #422, opened 2019-09-14
Use tue-apt-select-mirror in CI

MatthijsBurgh:


setup/tue-functions.bash
```diff
@@ -32,8 +32,8 @@ function tue-apt-select-mirror
     # It uses apt-select to generate a new sources.list, based on the current one.
     # All Arguments to this functions are passed on to apt-select, so check the
     # apt-select documentation for all options.
-	hash pip2 2> /dev/null|| sudo apt-get install python-pip
-	hash apt-select 2> /dev/null|| pip2 install --user apt-select
+	hash pip2 2> /dev/null|| sudo apt-get install --assume-yes python-pip
+	hash apt-select 2> /dev/null|| sudo -H pip2 install apt-select
```
ar13pit: Why the change to a system-wide install?

MatthijsBurgh: With local install, I get a 127, command not found error.

ar13pit: That shouldn't happen. On my PC it works completely fine with local install

MatthijsBurgh: Yep also on my PC, but not in CI

ar13pit: So not in Docker you mean?

MatthijsBurgh: Yep, not in docker.

PR #426, opened 2019-10-09
Commit range to determine changed packages

MatthijsBurgh:
Right now the CI only checks the modified packages against the last commit in a push. (in a PR, all commits are checked.) If you don't push every commit separately, these commit are put together in one build. But the still only the last commit was used to check for changed packages.

If this PR, this is fixed. For Travis, there is already a variable available. For azure, a script is needed to generate this. A external API is used to determine this. Azure is inconsistent and sometimes only provides one commit, while multiple commits were added since the last push. Therefore the script returns an empty string the case of just one commit. `Packages.sh` falls back to old implementation.

Update: Azure limits the number of commits to 200. In that case, the old method would be better.

### DIFF
--- a/ci/.travis.yml.multiple_pkgs.example
+++ b/ci/.travis.yml.multiple_pkgs.example
@@ -11,7 +11,7 @@ before_install:
   - wget https://raw.githubusercontent.com/tue-robotics/tue-env/master/ci/packages.sh
   # Determine packages modified in this commit
   - ALL=$([ $TRAVIS_EVENT_TYPE == "cron" ] && echo "true")
-  - source packages.sh --pullrequest=$TRAVIS_PULL_REQUEST --branch=$TRAVIS_BRANCH --all=$ALL # add all folders which need to be excluded as an argument
+  - source packages.sh --commit-range=$TRAVIS_COMMIT_RANGE --all=$ALL # add all folders which need to be excluded as an argument
 
 env:
   # Run following in root of repo:
--- a/ci/azure_commit_range.py
+++ b/ci/azure_commit_range.py
@@ -34,10 +34,3 @@ newest_commit = json_data["value"][0]["id"]
 oldest_commit = json_data["value"][number_commits-1]["id"]
 commit_range = "{}...{}".format(newest_commit, oldest_commit)
 print(commit_range)
-
-
-
-
-
-
-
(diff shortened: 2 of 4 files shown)


ci/azure_commit_range.py
```diff
@@ -0,0 +1,43 @@
...
+oldest_commit = json_data["value"][number_commits-1]["id"]
+commit_range = "{}...{}".format(newest_commit, oldest_commit)
+print(commit_range)
+
+
+
+
+
+
+
```
ar13pit: Blank lines

ci/.travis.yml.multiple_pkgs.example
```diff
@@ -11,7 +11,7 @@ before_install:
   - wget https://raw.githubusercontent.com/tue-robotics/tue-env/master/ci/packages.sh
   # Determine packages modified in this commit
   - ALL=$([ $TRAVIS_EVENT_TYPE == "cron" ] && echo "true")
-  - source packages.sh --pullrequest=$TRAVIS_PULL_REQUEST --branch=$TRAVIS_BRANCH --all=$ALL # add all folders which need to be excluded as an argument
+  - source packages.sh --commit-range=$TRAVIS_COMMIT_RANGE --all=$ALL # add all folders which need to be excluded as an argument
```
ar13pit: Is `pullrequest` arg not required for travis?

MatthijsBurgh: Nope, as all the information is now included in the commit range. Or could you subscribe a situation in which you think is is required?

ar13pit: Is there a way to make this better so that for both azure and travis the syntax is the same? Because azure is requiring pullrequest arg and travis not.

MatthijsBurgh: The pullrequest argument for azure should only be temporally, As the commits assigned to a build is inconsistent. Recently, it looks to be consistent, but more time is needed to be sure.

MatthijsBurgh: Small remark, the azure rest API only allows for max 200 commits, so this should be taken into account.

ar13pit: Alright. This PR will require a part 2 sometime soon.

PR #440, opened 2019-12-23
melodic-devel

MatthijsBurgh:


Dockerfile
```diff
@@ -25,6 +25,9 @@ ENV DEBIAN_FRONTEND=noninteractive \
 # Set default shell to be bash
 SHELL ["/bin/bash", "-c"]
 
+# Set Timezone for tzdata
+RUN ln -fs /usr/share/zoneinfo/Europe/Amsterdam /etc/localtime
```
ar13pit: This is not required anymore.

ar13pit:
@MatthijsBurgh I think we can close this PR for now. The changes are only in base docker images. When we move to melodic, this can be done again :).

PR #443, opened 2019-12-29
Fixes to use tue-get on multiple users simultaneous

MatthijsBurgh:
- Fix acces rights for all users to state dir in /tmp folder
- Make tmp files user and session unique
- Add more debug prints
- Add --purge autocomplete for `tue-env remove`
- Add some separation lines


installer/tue-install-impl.bash
```diff
@@ -839,6 +886,8 @@ do
     fi
 done
 
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
```
ar13pit: Why these here? The script was pretty clear without these separators. This is just making it ugly.

MatthijsBurgh: My opinion was that the main loop needed a better structure. I converted the separation lines to double empty lines.

installer/tue-install-impl.bash
```diff
@@ -777,7 +835,16 @@ ROS_PACKAGE_INSTALL_DIR=$TUE_SYSTEM_DIR/src
 
 TUE_INSTALL_SCRIPTS_DIR=$TUE_DIR/installer
 
-TUE_INSTALL_STATE_DIR=/tmp/tue-installer/$stamp
+TUE_INSTALL_GENERAL_STATE_DIR=/tmp/tue-installer
+if [ ! -d $TUE_INSTALL_GENERAL_STATE_DIR ]
+then
+    tue-install-debug "mkdir $TUE_INSTALL_GENERAL_STATE_DIR"
+    mkdir "$TUE_INSTALL_GENERAL_STATE_DIR"
```
Rayman: You could use the `mktemp` utility to safely create a temporary directory with a prefix.

Issue #369, opened 2019-02-14
Uninstalling target doesn't stop sourcing of setup


MatthijsBurgh:
Because the target file in the dependencies(-on) isn't removed, it's setup file is still sourced. This is a significant bug!

ar13pit:
Well if the target is removed then the setup file doesn't exist anymore so what exactly is the issue?

MatthijsBurgh:
Uninstalling, Title wasn't clear

ar13pit:
I think I know how to solve this issue. The problem is in `tue-get` function of `tue-functions.bash` and not `tue-install-impl.bash`.

In the removal process we just go to the `installed` directory and delete the `target` file. What we must also do is first check if there exists any `target` file in `dependencies-on`. If exists, then do nothing. 

If does not exist then the following are the options:
1. Delete the `target` file in `dependencies` and run `tue-get update`. This will remove the setup file of `target` from `targets_setup.bash` but all the dependencies of `target` will still be in the workspace even though the only target they have dependencies on was `target`

2. Recursively read the `target` file in `dependencies` and for every dependency `target_parent` check if inside the `dependencies-on/$target_parent` there is something more than 'target'. If not then remove `$target_parent`. This will make sure that `target_parents` that have no children will be removed from the environment. 

In either of these options, the only thing that will be removed is the sourcing of packages in the environment and not the packages themselves.

@MatthijsBurgh @Timple is my analogy correct or am I missing something?

MatthijsBurgh:
I ~don't like~ hate option 1.

You can only remove a target if not in list_installed and empty dependencies-on. Empty dependencies-on is not a guarantee.

ar13pit:
Are you suggesting that if the `target` exists in `installed` and a removal command is issued, we also remove the children of this target that are listed in `dependencies-on/target` and then go for option 1? Because if we go for option 1 before removing the children, a lot of packages will break.

To summarize we have the following options:
**If a target exists in the installed list then** a removal has the following possibilities:
**If the target has no children then**
1. Only remove the target and not its single parents
2. Remove the target and its single parents as well

**If the target has children then**
3. Cancel the removal
4. Remove the children and then only remove the target
5. Remove the children, the target and the single parents of the target

@MatthijsBurgh You get to choose one option from (1, 2) and one from (3, 4, 5)

MatthijsBurgh:
Fixed by #446

PR #446, opened 2020-01-25
Much better remove

MatthijsBurgh:
We could move the `generate_setup_file` from `tue_install_impl.bash`, so that the setup file could be regenerated after a successful removal.


ar13pit:
Does this resolve #369 ?

MatthijsBurgh:
Yes, no more sourcing of setup files if a target is removed.

ar13pit:
But you still need to move `generate_setup_file`. Without that removal is incomplete as we would need to run `tue-get update` after the current proposal to fix the setup file

MatthijsBurgh:
> But you still need to move `generate_setup_file`. Without that removal is incomplete as we would need to run `tue-get update` after the current proposal to fix the setup file

Indeed. Right now `tue-get update`  is required to generate the setup file. So I will move this.

installer/tue-install-impl.bash
```diff
@@ -879,16 +843,7 @@ done
...
-echo "# This file was auto-generated by tue-install. Do not change this file." > "$TUE_ENV_DIR"/.env/setup/target_setup.bash
-
-mkdir -p "$TUE_INSTALL_DEPENDENCIES_DIR"
-installed_targets=$(ls "$TUE_INSTALL_DEPENDENCIES_DIR")
-TUE_SETUP_TARGETS=" "
-for t in $installed_targets
-do
-    generate_setup_file "$t"
-done
+_generate_setup_file
```
ar13pit: I think this can be moved completely to tue-functions.bash as after tue-install-impl exits, the stack goes back to `tue-get`

MatthijsBurgh: Do we want it happen before or after all the installs?

ar13pit: I would say after. Then the setup file would be the right one. And if install/update fails then the setup file doesn't change.

installer/tue-install-impl.bash
```diff
@@ -1019,3 +970,5 @@ then
         done
     fi
 fi
+
+return 0
```
ar13pit: This is actually not required as nothing is checking this return value.

setup/tue-functions.bash
```diff
@@ -572,6 +572,129 @@ function _show_file
...
+}
+
+function _generate_setup_file
+{
+    mkdir -p "$TUE_ENV_DIR"/.env/setup
+    echo "# This file was auto-generated by tue-get. Do not change this file." > "$TUE_ENV_DIR"/.env/setup/target_setup.bash
+
+    tue_dependencies_dir="$TUE_ENV_DIR"/.env/dependencies
+
+    mkdir -p "$tue_dependencies_dir"
```
ar13pit: Why create this directory? Just execute the statements below only if the directory exists.

setup/tue-functions.bash
```diff
@@ -572,6 +572,129 @@ function _show_file
...
+    local installed_targets
+    installed_targets=$(ls "$tue_dependencies_dir")
+    TUE_SETUP_TARGETS=" "
+    for t in $installed_targets
+    do
+        __generate_setup_file "$t"
+    done
+    unset TUE_SETUP_TARGETS
+    unset tue_dependencies_dir
```
ar13pit: This is not required.

setup/tue-functions.bash
```diff
@@ -572,6 +572,129 @@ function _show_file
...
+
+    tue_dependencies_dir="$TUE_ENV_DIR"/.env/dependencies
+
+    mkdir -p "$tue_dependencies_dir"
+    local installed_targets
+    installed_targets=$(ls "$tue_dependencies_dir")
+    TUE_SETUP_TARGETS=" "
+    for t in $installed_targets
+    do
+        __generate_setup_file "$t"
```
ar13pit: The function can be removed as it is only called here. Instead condense the code and put it in the for loop. The logic becomes more readable.

ar13pit: Also this would eliminate the need for passing a variable as a global variable to the function.

setup/tue-functions.bash
MatthijsBurgh: Why not? They will stay alive, clean env is better

setup/tue-functions.bash
MatthijsBurgh: Whatever you like

setup/tue-functions.bash
MatthijsBurgh: This function is called recursive...........

installer/tue-install-impl.bash
MatthijsBurgh: Yep it is, in the tue-functions.bash

setup/tue-functions.bash
MatthijsBurgh: Are local variables available in the functions called?

MatthijsBurgh: Put it in the for loop? Which?

installer/tue-install-impl.bash
ar13pit: But this return from `tue-install-impl` to `tue-install`. The return of `tue-install` is checked in `tue-functions`.
Am I right or is it something that I'm not seeing here?

MatthijsBurgh: Yep, I have tested it and the return of `tue-install-impl` does end up in `tue-functions`. `tue-install` can't return as it is not sourced, but run in a child process.

ar13pit: Alright.

setup/tue-functions.bash
MatthijsBurgh: These can be local variables.

setup/tue-functions.bash
MatthijsBurgh: Both `__generate_setup_file ` and `_generate_setup_file ` are needed. `__generate_setup_file ` is called recursively. `_generate_setup_file ` is called on two spots

PR #449, opened 2020-02-01
Nobleo backport

Timple:
Performed a two-way-sync with nobleo-get :). Contains some new options mostly

### DIFF
--- a/installer/bootstrap.bash
+++ b/installer/bootstrap.bash
@@ -1,9 +1,9 @@
 #! /usr/bin/env bash
 
 # Make sure git is installed
-hash git 2> /dev/null || sudo apt-get install --assume-yes --no-install-recommends git
+hash git 2> /dev/null || sudo apt-get install --assume-yes -qq --no-install-recommends git
 # Make sure lsb-release is installed
-hash lsb_release 2> /dev/null || sudo apt-get install --assume-yes lsb-release
+hash lsb_release 2> /dev/null || sudo apt-get install --assume-yes -qq lsb-release
 
 # Check if OS is Ubuntu
 # shellcheck disable=SC1091
--- a/installer/bootstrap.bash
+++ b/installer/bootstrap.bash
@@ -1,7 +1,7 @@
 #! /usr/bin/env bash
 
 # Make sure git is installed
-hash git 2> /dev/null || sudo apt-get install --assume-yes --no-install-recommends git
+hash git 2> /dev/null || sudo apt-get install --assume-yes git
 # Make sure lsb-release is installed
 hash lsb_release 2> /dev/null || sudo apt-get install --assume-yes lsb-release
 
--- a/installer/bootstrap.bash
+++ b/installer/bootstrap.bash
@@ -1,7 +1,7 @@
 #! /usr/bin/env bash
 
 # Make sure git is installed
-hash git 2> /dev/null || sudo apt-get install --assume-yes git
+hash git 2> /dev/null || sudo apt-get install --assume-yes --no-install-recommends git
 # Make sure lsb-release is installed
 hash lsb_release 2> /dev/null || sudo apt-get install --assume-yes lsb-release
 
--- a/installer/parse-ros-package-deps.py
+++ b/installer/parse-ros-package-deps.py
@@ -35,10 +35,11 @@ def main():
     deps = doc.findall('exec_depend')
     dep_set |= {dep.text for dep in deps}
 
-    if os.getenv('ROBOT_REAL', 'false') != 'true':  # Skip test and doc dependencies on actual bots
+    if os.getenv('TUE_INSTALL_TEST_DEPEND', 'false') == 'true':
         deps = doc.findall('test_depend')
         dep_set |= {dep.text for dep in deps}
 
+    if os.getenv('TUE_INSTALL_DOC_DEPEND', 'false') == 'true':
         deps = doc.findall('doc_depend')
         dep_set |= {dep.text for dep in deps}
 
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -34,7 +34,7 @@ function tue-install-error
 Error while installing target '$TUE_INSTALL_CURRENT_TARGET':
 
     $1
-\033[0m" | tee --append $INSTALL_DETAILS_FILE
+\033[0m" | tee --append "$INSTALL_DETAILS_FILE"
     exit 1
 }
 
@@ -532,6 +532,7 @@ function tue-install-system-now
 
         # shellcheck disable=SC2086
         sudo apt-get install --assume-yes -qq --no-install-recommends $pkgs_to_install || sudo apt-get update  # Update and try again
+        # shellcheck disable=SC2086
         sudo apt-get install --assume-yes -qq --no-install-recommends $pkgs_to_install || exit 1
         tue-install-debug "Installed $pkgs_to_install ($?)"
     fi
@@ -610,10 +611,14 @@ function tue-install-snap-now
 
     tue-install-system-now snapd
 
+
+    local snaps_to_install snaps_installed
     snaps_to_install=""
     snaps_installed=$(snap list)
+    # shellcheck disable=SC2068
     for pkg in $@
     do
+        echo "$pkg"
         if [[ ! $snaps_installed == *$pkg* ]] # Check if pkg is not already installed
         then
             snaps_to_install="$snaps_to_install $pkg"
@@ -630,7 +635,7 @@ function tue-install-snap-now
         do
             echo -e "yes | sudo snap install --classic $pkg\n"
             tue-install-debug "yes | sudo snap install --classic $pkg"
-            yes | sudo snap install --classic $pkg
+            yes | sudo snap install --classic "$pkg"
         done
     fi
 }
@@ -922,7 +927,7 @@ then
     for ppa in ${TUE_INSTALL_PPA}
     do
         ppa=${ppa//#/ }  # Replace # back to spaces
-        if [ -z "$(grep -h "^deb.*${ppa#ppa:}" /etc/apt/sources.list.d/* 2>&1)" ] && ! grep -q "$ppa" /etc/apt/sources.list
+        if ! grep -h "^deb.*${ppa#ppa:}" /etc/apt/sources.list.d/* 1>/dev/null 2>&1 && ! grep -q "$ppa" /etc/apt/sources.list
         then
             tue-install-system-now software-properties-common
             tue-install-info "Adding ppa: $ppa"
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -912,7 +912,7 @@ function tue-install-ros
     local ros_pkg_name=${TUE_INSTALL_CURRENT_TARGET#ros-}
     if [[ $ros_pkg_name == *-* ]]
     then
-        tue-install-error "A ROS package cannot contain dashes (${ros_pkg_name}), rename to ros-${ros_pkg_name//-/_}"
+        tue-install-error "A ROS package cannot contain dashes (${ros_pkg_name}), make sure the package is named '${ros_pkg_name//-/_}' and rename the target to 'ros-${ros_pkg_name//-/_}'"
         return 1
     fi
 
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -813,11 +813,15 @@ function _tue-install-pip-now
     fi
 
     local pips_to_install=""
+    local git_pips_to_install=""
     # shellcheck disable=SC2048
     for pkg in $*
     do
         local installed_version
-        if [[ "$pkg" == "git+"* ]] || ! installed_version=$(python"${pv}" "$TUE_INSTALL_SCRIPTS_DIR"/check-pip-pkg-installed-version.py "$pkg")
+        if [[ "$pkg" == "git+"* ]]
+        then
+            git_pips_to_install="$git_pips_to_install $pkg"
+        elif ! installed_version=$(python"${pv}" "$TUE_INSTALL_SCRIPTS_DIR"/check-pip-pkg-installed-version.py "$pkg")
         then
             pips_to_install="$pips_to_install $pkg"
         else
@@ -831,6 +835,17 @@ function _tue-install-pip-now
         echo -e "yes | pip${pv} install --user $pips_to_install\n"
         # shellcheck disable=SC2048,SC2086
         yes | pip"${pv}" install --user $pips_to_install
+    fi
+
+    if [ -n "$git_pips_to_install" ]
+    then
+        for pkg in $git_pips_to_install
+        do
+            echo -e "Going to run the following command:\n"
+            echo -e "yes | pip${pv} install --user $pkg\n"
+            # shellcheck disable=SC2048,SC2086
+            yes | pip"${pv}" install --user $pkg
+        done
     fi
 }
 
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -742,6 +742,19 @@ function _tue-install-pip-now
         tue-install-error "Invalid tue-install-pip${pv}-now call: needs package as argument."
     fi
 
+    # Make sure pip is up-to-date before checking version and installing
+    local pip_version desired_pip_version
+    pip_version=$(pip"${pv}" --version | awk '{print $2}')
+    desired_pip_version="20"
+    if version_gt "$desired_pip_version" "$pip_version"
+    then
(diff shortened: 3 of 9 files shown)


installer/parse-ros-package-deps.py
```diff
@@ -29,11 +33,12 @@ def main():
     deps = doc.findall('exec_depend')
     dep_set |= set([dep.text for dep in deps])
 
-    deps = doc.findall('test_depend')
-    dep_set |= set([dep.text for dep in deps])
+    if os.getenv('ROBOT_REAL', 'false') != 'true':  # Skip test and doc dependencies on actual bots
```
MatthijsBurgh: But ROBOT_REAL will not be set, when running the installer for the first time.

Timple: Ah yes, would it be wise to implement something that does work? 
A proper solution doesn't really come to mind at the moment..

ar13pit: If we export  `ROBOT_REAL` in `install.bash` of the robot target, will this work then?

MatthijsBurgh: No, as it is not the first target to be installed.

MatthijsBurgh: I think we shouldn't install these by default, unless you provide an extra argument or config item.

ar13pit: That's sounds like a better option.

Timple: That could work.
In that case I would make the config item default if the 'user' target is installed? Not testing in developer mode might encourage bad practice?

MatthijsBurgh:
@ar13pit I think this one is good to go. Please review

installer/tue-install-impl.bash
```diff
@@ -875,14 +903,18 @@ function tue-install-ros
     [ -n "$TUE_ROS_DISTRO" ] || tue-install-error "Environment variable 'TUE_ROS_DISTRO' is not set."
 
     local ros_pkg_name=${TUE_INSTALL_CURRENT_TARGET#ros-}
+    if [[ $ros_pkg_name == *-* ]]
+    then
+        tue-install-error "A ROS package cannot contain dashes (${ros_pkg_name})"
```
ar13pit: Update this error message. 
```suggestion
        tue-install-error "A ROS package cannot contain dashes (${ros_pkg_name}). Fix target by replacing '-' with '_'."
```

installer/tue-install-impl.bash
```diff
@@ -1018,6 +1055,24 @@ do
...
+        TUE_INSTALL_DOC_DEPEND=false
+            ;;
+        --test-depend)
+        # shellcheck disable=SC2034
+        TUE_INSTALL_TEST_DEPEND=true
+            ;;
+        --no-test-depend)
+        # shellcheck disable=SC2034
+        TUE_INSTALL_TEST_DEPEND=false
+            ;;
```
ar13pit: Maybe set the default value of TUE_INSTALL_TEST_DEPEND and TUE_INSTALL_DOC_DEPEND to be false. If the option is used then the value is set as true.

setup/tue-functions.bash
```diff
@@ -709,16 +717,21 @@ function tue-get
...
 
     Possible options:
-        --debug        - Shows more debugging information
-        --branch=name  - Try to checkout this branch if exists
+        --debug          - Shows more debugging information
+        --no-ros-deps    - Do not install ROS dependencies
+        --doc-depend     - Do install doc dependencies, overules config
+        --no-doc-depend  - Do not install doc dependencies, overules config
+        --test-depend    - Do install test dependencies, overules config
+        --no-test-depend - Do not install test dependencies, overules config
```
ar13pit: These need to be updated accordingly.

setup/tue-functions.bash
```diff
@@ -923,17 +936,17 @@ function _tue-get
...
-            mapfile -t COMPREPLY < <(compgen -W "$(echo -e "$(find "$TUE_ENV_DIR"/.env/dependencies -mindepth 1 -maxdepth 1 -type f -not -name ".*" -printf "%f\n" | sed "s/.*/'& '/g")\n'--plain '\n'--verbose '\n'--all '\n'--level='")" -- "$cur")
+            mapfile -t COMPREPLY < <(compgen -W "$(echo -e "$(find "$TUE_ENV_DIR"/.env/dependencies -mindepth 1 -maxdepth 1 -type f -not -name ".*" -printf "%f\n" | sed "s/.*/'& '/g")\n'--plain '\n'--verbose '\n'--ros-only '\n'--all '\n'--level='")" -- "$cur")
         elif [[ $cmd == "update" ]]
         then
             local IFS=$'\n'
             # shellcheck disable=SC2178
-            mapfile -t COMPREPLY < <(compgen -W "$(echo -e "$(find "$TUE_ENV_DIR"/.env/dependencies -mindepth 1 -maxdepth 1 -type f -not -name ".*" -printf "%f\n" | sed "s/.*/'& '/g")\n'--debug '\n'--branch='")" -- "$cur")
+            mapfile -t COMPREPLY < <(compgen -W "$(echo -e "$(find "$TUE_ENV_DIR"/.env/dependencies -mindepth 1 -maxdepth 1 -type f -not -name ".*" -printf "%f\n" | sed "s/.*/'& '/g")\n'--debug '\n'--no-ros-deps '\n'--doc-depend '\n'--no-doc-depend '\n'--test-depend '\n'--no-test-depend '\n'--branch='")" -- "$cur")
```
ar13pit: These as well

installer/tue-install-impl.bash
```diff
@@ -640,7 +658,9 @@ function tue-install-system-now
         done
 
         # shellcheck disable=SC2086
-        sudo apt-get install --assume-yes $pkgs_to_install
+        sudo apt-get install --assume-yes -qq --no-install-recommends $pkgs_to_install || sudo apt-get update  # Update and try again
+        # shellcheck disable=SC2086
+        sudo apt-get install --assume-yes -qq --no-install-recommends $pkgs_to_install || exit 1
```
ar13pit: By default run update first and the install. Instead of exit, use tue-install-error.
```suggestion
        sudo apt-get update -qq 
        # shellcheck disable=SC2086
        sudo apt-get install --assume-yes -qq --no-install-recommends $pkgs_to_install || tue-install-error "An error occurred in installing system packages."
```

installer/tue-install-impl.bash
```diff
@@ -655,13 +675,14 @@ function tue-install-ppa
     then
         tue-install-error "Invalid tue-install-ppa call: needs ppa as argument."
     fi
+    local ppa="$*"
 
-    if [[ ! $1 == ppa:* ]]
+    if [[ $ppa != ppa:* && $ppa != deb* ]]
```
ar13pit: This needs to be added in the docs that `deb` sources can now be installed using `tue-install-ppa`

installer/tue-install-impl.bash
```diff
@@ -670,14 +691,21 @@ function tue-install-ppa-now
...
+        tue-install-error "Invalid tue-install-ppa-now call: needs ppa as argument."
+    fi
+
     local PPA_ADDED=""
     # shellcheck disable=SC2048
     for ppa in $*
     do
-        if ! grep -h "^deb.*${ppa#ppa:}" /etc/apt/sources.list.d/* 1>/dev/null 2>&1
+        ppa="${ppa//^/ }"
+        if ! grep -q -h "^deb.*${ppa#ppa:}" /etc/apt/sources.list.d/* 2>&1
```
ar13pit: What about the `deb` source ppa? Are you sure the variable substitution will work for ppas with `deb <src>`?

installer/tue-install-impl.bash
```diff
@@ -204,13 +213,19 @@ function _try_branch_git
     fi
 
     tue-install-debug "git -C $1 checkout $2"
-    local _try_branch_res
```
ar13pit: Add a comment that this is a global variable

installer/tue-install-impl.bash
```diff
@@ -252,34 +267,36 @@ function tue-install-git
...
         _try_branch_git "$targetdir" "$version"
-        res="$res $_try_branch_res"
+        [ -n "$_try_branch_res" ] && res="$res $_try_branch_res"
     fi
 
     tue-install-debug "Desired branch: $BRANCH"
-    if [ -n "$BRANCH" ]; #Cannot be combined with version-if because this one might not exist
+    if [ -n "$BRANCH" ] #Cannot be combined with version-if because this one might not exist
     then
+        local _try_branch_res
```
ar13pit: Define this only once. Clear it in the `if` block if needed

installer/tue-install-impl.bash
```diff
@@ -428,7 +445,8 @@ function tue-install-cp
 
     if [ -z "$2" ]
     then
-        tue-install-error "Invalid tue-install-cp call: needs two arguments (source and target). The source must be relative to the installer target directory"
+        tue-install-error "Invalid tue-install-cp call: needs two arguments (source and target). The source must be relative to the installer target directory
+        Command: tue-install-cp $*"
```
ar13pit: This will be printed with an indent

setup/tue-env-config.bash
```diff
@@ -36,6 +36,42 @@ function tue-env-use-https
...
+
+function tue-env-not-install-doc-depend
+{
+    local option="TUE_INSTALL_DOC_DEPEND"
+    local value="false"
+    _set_export_option "$option" "$value" "$tue_env_dir"/.env/setup/user_setup.bash
+
+    echo -e "[tue-env](config) Environment '$env' set to not install doc dependencies"
+}
+
```
ar13pit: I need to check this up with my modified fork. I had some nice modifications for this kind of functionality.

installer/parse-ros-package-deps.py
```diff
@@ -31,11 +35,13 @@ def main():
     deps = doc.findall('exec_depend')
     dep_set |= {dep.text for dep in deps}
 
-    deps = doc.findall('test_depend')
-    dep_set |= {dep.text for dep in deps}
+    if os.getenv('TUE_INSTALL_TEST_DEPEND', 'false') == 'true':
```
ar13pit: If this environment variable is being modified in tue-install-impl but not exported with the flag --doc-depend, will the `if` condition resolve correctly?

installer/bootstrap.bash
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env bash
 
 # Make sure git is installed
-hash git 2> /dev/null || sudo apt-get install --assume-yes git
+hash git 2> /dev/null || sudo apt-get install --assume-yes --no-install-recommends git
 # Make sure lsb-release is installed
 hash lsb_release 2> /dev/null || sudo apt-get install --assume-yes lsb-release
```
ar13pit:
```suggestion
hash git 2> /dev/null || sudo apt-get install --assume-yes -qq --no-install-recommends git
# Make sure lsb-release is installed
hash lsb_release 2> /dev/null || sudo apt-get install --assume-yes -qq lsb-release
```

installer/tue-install-impl.bash
```diff
@@ -855,8 +883,8 @@ function tue-install-dpkg
     fi
     tue-install-debug "Installing dpkg $1"
     sudo dpkg --install "$1"
-    tue-install-debug "sudo apt-get --fix-broken --assume-yes install"
-    sudo apt-get --fix-broken --assume-yes install
+    tue-install-debug "sudo apt-get --fix-broken --assume-yes --no-install-recommends install"
+    sudo apt-get --fix-broken --assume-yes  --no-install-recommends install
```
ar13pit:
```suggestion
    tue-install-debug "sudo apt-get --fix-broken --assume-yes -qq --no-install-recommends install"
    sudo apt-get --fix-broken --assume-yes -qq --no-install-recommends install
```

installer/tue-install-impl.bash
MatthijsBurgh: The default is what you put in your config

ar13pit: That's a bit messy in my opinion. We should keep the defaults here and overrides in the config. As a typical user will overlook this config. The ones who have the knowledge about how tue-env works will use the config file correctly.

installer/tue-install-impl.bash
```diff
@@ -14,6 +14,15 @@ TUE_REPOS_DIR=$TUE_ENV_DIR/repos
 
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
+function _skip_in_ci
+{
+    # use as followed:
+    # _skip_in_ci || return 0
```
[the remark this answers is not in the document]
Timple: Stupid new github layout. I've commented this way earlier but I always have this stuff pending on some button...

installer/tue-install-impl.bash
Timple: Nope, this was done for a speedup. An `apt-get update` is not often required (typically happens even daily in the background on most laptops. So this will save time 90% of the time costing a little extra on the 10%.
Also running `nobleo-get install` with different options multiple times will be quicker

installer/tue-install-impl.bash
Timple: This is a misleading suggestion. People who don't know what the error is about and need this suggestion will create `ros_std_msgs` and it still won't work.

Timple:
Thanks both for the effort of updating btw :dancing_men:

installer/tue-install-impl.bash
MatthijsBurgh: If you don't have a config, the default is false.

installer/tue-install-impl.bash
MatthijsBurgh: Now provides the new name.

installer/tue-install-impl.bash
MatthijsBurgh: Changed so update is done and tmp file is created, so no new update is done.

installer/tue-install-impl.bash
ar13pit: If there exists no config file then these options won't work as the variables are not exported.

installer/tue-install-impl.bash
```diff
@@ -875,14 +910,18 @@ function tue-install-ros
     [ -n "$TUE_ROS_DISTRO" ] || tue-install-error "Environment variable 'TUE_ROS_DISTRO' is not set."
 
     local ros_pkg_name=${TUE_INSTALL_CURRENT_TARGET#ros-}
+    if [[ $ros_pkg_name == *-* ]]
+    then
+        tue-install-error "A ROS package cannot contain dashes (${ros_pkg_name}), rename to ros-${ros_pkg_name//-/_}"
```
ar13pit: With the word package, @Timple 's argument still holds.
```suggestion
        tue-install-error "A ROS target cannot contain dashes (${ros_pkg_name}), rename to ros-${ros_pkg_name//-/_}"
```

installer/tue-install-impl.bash
```diff
@@ -639,8 +657,17 @@ function tue-install-system-now
...
+        then
+            # Update once every boot. Or delete the tmp file if you need an update before installing a pkg.
+            tue-install-debug "sudo apt-get update -qq"
+            sudo apt-get update -qq
+            /tmp/apt_get_update
+        fi
+
         # shellcheck disable=SC2086
-        sudo apt-get install --assume-yes $pkgs_to_install
+        sudo apt-get install --assume-yes -qq --no-install-recommends $pkgs_to_install || exit 1
```
ar13pit:
```suggestion
        sudo apt-get install --assume-yes -qq --no-install-recommends $pkgs_to_install || tue-install-error "An error occurred while installing system packages."
```

installer/parse-ros-package-deps.py
ar13pit: See updated comment in the options of tue-install-impl.

installer/tue-install-impl.bash
MatthijsBurgh: The debs will be added tot /etc/apt/sources.list. So this will not work

installer/tue-install-impl.bash
MatthijsBurgh: Ok, then we export them.

installer/tue-install-impl.bash
MatthijsBurgh: See new commit

installer/tue-install-impl.bash
MatthijsBurgh: Done

installer/tue-install-impl.bash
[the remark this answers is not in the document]
MatthijsBurgh: changed so it should used like `_skip_in_ci && return 0`

MatthijsBurgh:
@ar13pit export in subprocess doesn't affect the parent. So ready for review.

PR #453, opened 2020-02-05
Refactor

MatthijsBurgh:
- more `tue-install-XXX-now`
- first version of mercurial support
- small fixes
- small cleanup
- use python3 in scripts
- extend yaml scheme


MatthijsBurgh:
tue-install-pip-now-filtered probably doesn't work if version constraints are provided.

installer/tue-install-impl.bash
```diff
@@ -806,17 +1036,44 @@ TUE_INSTALL_STATE_DIR=$TUE_INSTALL_GENERAL_STATE_DIR/$stamp
...
 TUE_INSTALL_PIP3S=
 TUE_INSTALL_SNAPS=
 
 TUE_INSTALL_WARNINGS=
 TUE_INSTALL_INFOS=
 
+# Make sure tools used by this installer are installed
+tue-install-system-now git python-pip python3-pip
+
+tue-install-pip2-now mercurial
```
ar13pit: Why not system-now? If really pip then why not pip3?

installer/tue-install-impl.bash
```diff
@@ -806,17 +1036,44 @@ TUE_INSTALL_STATE_DIR=$TUE_INSTALL_GENERAL_STATE_DIR/$stamp
...
+tue-install-pip3-now distro PyYAML
+
+# Mercurial config extension to write configs from cli
+hgcfg_folder="$HOME"/src/hgcfg
+hgcfg_pulled=/tmp/hgcfg_pulled
+if [ ! -f "$hgcfg_pulled" ]
+then
+    parent_target=TUE_INSTALL_CURRENT_TARGET
+    TUE_INSTALL_CURRENT_TARGET="hgcfg"
+    tue-install-git "https://github.com/tue-robotics/hgconfig.git" "$hgcfg_folder"
```
ar13pit: I don't like this. If possible move this to the targets repo and try to generalize this entire if block by using default config if none exists.

installer/tue-install-impl.bash
```diff
@@ -556,28 +648,127 @@ function tue-install-ppa
...
-        tue-install-error "Invalid tue-install-pip3 call: needs package as argument."
+        tue-install-error "Invalid tue-install-pip${pv}-now call: needs package as argument."
+    fi
+
+    local pips_to_install=""
+    # shellcheck disable=SC2048
+    for pkg in $*
+    do
+        local installed_version
+        if [[ "$pkg" == "git+"* ]] || ! installed_version=$(python"${pv}" "$TUE_INSTALL_SCRIPTS_DIR"/pip-correctly-installed.py "$pkg")
```
ar13pit: Maybe rename the script `pip-correctly-installed` to something like `check-pip-pkg-installed-version` because right now it sounds more like its job is to check the installation of pip itself rather than the package.

installer/tue-install-impl.bash
MatthijsBurgh: This hg extension is required to edit the config in the command line. By default, hg opens an editor to edit the config.

The config contains the url, which we want to change if different.

installer/tue-install-impl.bash
MatthijsBurgh: Apt is very old, 3.7. Pip is on 5.3. But the hgcfg extension is not yet compatible with python3.

installer/tue-install-impl.bash
MatthijsBurgh: BTW we could move this inside tue-install-hg

installer/tue-install-impl.bash
ar13pit: Okay. Then this remains unchanged for now.

installer/tue-install-impl.bash
```diff
@@ -270,20 +271,111 @@ function tue-install-git
...
+        if [[ $TUE_INSTALL_HG_PULL_Q =~ $targetdir ]]
+        then
+            tue-install-debug "Repo previously pulled, skipping"
+            # We have already pulled this repo, skip it
+            res=
+        else
+            # Switch url of origin to use https/ssh if different
+            # Get current remote url
+            local current_url
+            current_url=$(hg -R "$targetdir" cfg paths.default | awk '{print $2}')
```
ar13pit: I think there is a typo here. `cfg` -> `config`

installer/tue-install-impl.bash
ar13pit: I still don't see where exactly this hgconfig is required for using `tue-install-hg`. All the commands in `tue-install-hg` are executable without the requirement of any custom config. Can you pin point the exact line where this configuration is needed to execute `tue-install-hg`?

installer/tue-install-impl.bash
MatthijsBurgh: nope, `hgcfg` exposes `cfg`. I prefer to use `cfg`, so I am sure the `hgcfg` extension is used. (And error throws if there is something wrong with `hgcfg`) Which has a slightly different output.

installer/tue-install-impl.bash
MatthijsBurgh: https://github.com/tue-robotics/tue-env/pull/453/files#diff-27df6d20d171684702149fd4165c21e7R320, here we change the path. Without the extension. you can get items from the configs. But to set them from command line, without an editor, the extension is needed.

installer/tue-install-impl.bash
MatthijsBurgh: Rename done

ar13pit:
The CI is fine. I just changed a print statement so cancelled the build.

PR #467, opened 2020-03-09
Add more errors in case installing fails

MatthijsBurgh:
Weird branch name to test in action_server ci


MatthijsBurgh:
This causes the installer to stop. Which prevents situations in CI, where the installer continues, but the build fails for a weird reason.
https://travis-ci.com/tue-robotics/action_server/jobs/295971538#L2622 instead of https://travis-ci.com/tue-robotics/action_server/jobs/295947314#L2624

PR #470, opened 2020-03-11
Add ssh support for installing packages from private repos

ar13pit:
Also merge tue-robotics/hero_bringup#51 after this.


Dockerfile
```diff
@@ -28,16 +29,21 @@ SHELL ["/bin/bash", "-c"]
 # Install commands used in our scripts and standard present on a clean ubuntu
 # installation and setup a user with sudo priviledges
 RUN apt-get update -qq && \
-    apt-get install -qq --assume-yes --no-install-recommends apt-transport-https apt-utils ca-certificates curl dbus dialog git lsb-release sudo tzdata wget && \
+    apt-get install -qq --assume-yes --no-install-recommends apt-transport-https apt-utils ca-certificates curl dbus dialog git lsb-release openssh-client sudo tzdata wget > /dev/null && \
```
MatthijsBurgh: We have `-qq`. so why `> /dev/null`

.travis.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       language: generic
       services: docker
       name: tue-env
-      script: ci/build-docker-image.sh tue-env
+      script: ci/build-docker-image.sh -i="tuerobotics/tue-env" -b="$TRAVIS_BRANCH" -c="$TRAVIS_COMMIT" -pr="$TRAVIS_PULL_REQUEST" -prb="$TRAVIS_PULL_REQUEST_BRANCH" -u="$DOCKER_HUB_USERNAME" -p="$DOCKER_HUB_PASSWORD"
     - <<: *docker-generation
       name: tue-env-cuda
-      script: ci/build-docker-image.sh tue-env-cuda
+      script: ci/build-docker-image.sh -i="tuerobotics/tue-env-cuda" -b="$TRAVIS_BRANCH" -c="$TRAVIS_COMMIT" -pr="$TRAVIS_PULL_REQUEST" -prb="$TRAVIS_PULL_REQUEST_BRANCH" -u="$DOCKER_HUB_USERNAME" -p="$DOCKER_HUB_PASSWORD"
```
MatthijsBurgh: I prefer the long versions being use. It is longer but also way more readable.

Dockerfile
```diff
@@ -28,16 +29,21 @@ SHELL ["/bin/bash", "-c"]
...
+    usermod -aG adm $USER && \
+    echo "%sudo ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/"$USER"
 
 # Setup the current user and its home directory
 USER "$USER"
 WORKDIR /home/"$USER"
 
+ADD installer/bootstrap.bash ./bootstrap.bash
+
+RUN mkdir -p -m 0700 ~/.ssh && ssh-keyscan gitlab.com >> ~/.ssh/known_hosts && chmod 644 ~/.ssh/known_hosts
```
MatthijsBurgh: This should inherit from travis known hosts file

Dockerfile
```diff
@@ -28,16 +29,21 @@ SHELL ["/bin/bash", "-c"]
...
+    adduser -u 1000 --disabled-password --gecos "" $USER && \
+    usermod -aG sudo $USER && \
+    usermod -aG adm $USER && \
+    echo "%sudo ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/"$USER"
 
 # Setup the current user and its home directory
 USER "$USER"
 WORKDIR /home/"$USER"
 
+ADD installer/bootstrap.bash ./bootstrap.bash
```
MatthijsBurgh: I am not sure I prefer this over the old solution

ci/build-docker-image.sh
```diff
@@ -33,23 +63,30 @@ case $IMAGE_NAME_SUBSTRING in
...
+if [ "$CI_DOCKER_SSH" == "true" ]
+then
+    DOCKER_BUILDKIT=1 docker build --ssh=default --build-arg BRANCH="${CI_PULL_REQUEST_BRANCH:-$CI_BRANCH}" --build-arg \
+        PULL_REQUEST="$CI_PULL_REQUEST" --build-arg COMMIT="$CI_COMMIT" --build-arg \
+        CI="$CI" --build-arg BASE_IMAGE="$BASE_IMAGE" -t "$CI_DOCKER_IMAGE_NAME" .
+else
+    docker build --build-arg BRANCH="${CI_PULL_REQUEST_BRANCH:-$CI_BRANCH}" --build-arg \
+        PULL_REQUEST="$CI_PULL_REQUEST" --build-arg COMMIT="$CI_COMMIT" --build-arg \
+        CI="$CI" --build-arg BASE_IMAGE="$BASE_IMAGE" -t "$CI_DOCKER_IMAGE_NAME" .
+fi
```
MatthijsBurgh: Problem in using `DOCKER_BUILDKIT=1` when not using ssh?

ci/install-package.sh
```diff
@@ -29,6 +29,17 @@ do
...
             PULL_REQUEST="${i#*=}" ;;
+
+        -i=* | --image=* )
+            IMAGE_NAME="${i#*=}" ;;
+
+        -s=* | --ssh=* )
+            USE_SSH="${i#*=}"
+            USE_SSH=${USE_SSH:-"false"} ;;
+
+        -sk=* | --ssh-key=* )
```
MatthijsBurgh: `-sk` is against the arguments convention.

ci/install-package.sh
```diff
@@ -41,6 +52,7 @@ echo -e "\e[35m\e[1m PACKAGE      = ${PACKAGE} \e[0m"
 echo -e "\e[35m\e[1m BRANCH       = ${BRANCH} \e[0m"
 echo -e "\e[35m\e[1m COMMIT       = ${COMMIT} \e[0m"
 echo -e "\e[35m\e[1m PULL_REQUEST = ${PULL_REQUEST} \e[0m"
+echo -e "\e[35m\e[1m USE_SSH      = ${USE_SSH} \e[0m"
```
MatthijsBurgh: Maybe print the signature of the key

ci/install-package.sh
```diff
@@ -78,12 +88,18 @@ echo -e "\e[35m\e[1m Trying to fetch docker image: $IMAGE_NAME:$BRANCH_TAG \e[0m
...
-    BRANCH_TAG=master
+    docker pull "$IMAGE_NAME":"$MASTER_TAG"
+    BRANCH_TAG=$MASTER_TAG
 fi
 
 # Run the docker image along with setting new environment variables
-docker run --detach --interactive -e CI=true -e PACKAGE="$PACKAGE" -e BRANCH="$BRANCH" -e COMMIT="$COMMIT" -e PULL_REQUEST="$PULL_REQUEST" --name tue-env "$IMAGE_NAME:$BRANCH_TAG"
+docker run --detach --interactive -e CI="true" -e PACKAGE="$PACKAGE" -e BRANCH="$BRANCH" -e COMMIT="$COMMIT" -e PULL_REQUEST="$PULL_REQUEST" --name tue-env "$IMAGE_NAME:$BRANCH_TAG"
+
+if [ "$USE_SSH" != "false" ]
```
MatthijsBurgh:
```suggestion
if [ "$USE_SSH" == "true" ]
```
This is a better way as it forces people to do it right

.dockerignore
```diff
@@ -2,3 +2,4 @@
 #
 # Ignore the entire local repository
 /**
+!installer/bootstrap.bash
```
MatthijsBurgh: This should be revert if going back to old solution

ci/install-package.sh
ar13pit: Is it supposed to be just 1 letter argument?

Dockerfile
ar13pit: There's no difference between the two solutions. This would work in private gitlab instances too while the other would require a lot of configuration.

ci/install-package.sh
MatthijsBurgh: one dash should be followed by one letter, In case of multiple letters, multiple dashed should be used.

Dockerfile
ar13pit: To not pollute the CI output with things like
```
Selecting previously unselected package sgml-base.
(Reading database ... 11196 files and directories currently installed.)
Preparing to unpack .../sgml-base_1.26+nmu4ubuntu1_all.deb ...
Unpacking sgml-base (1.26+nmu4ubuntu1) ...
Selecting previously unselected package libpython2.7-minimal:amd64.
Preparing to unpack .../libpython2.7-minimal_2.7.12-1ubuntu0~16.04.9_amd64.deb ...
Unpacking libpython2.7-minimal:amd64 (2.7.12-1ubuntu0~16.04.9) ...
Selecting previously unselected package python2.7-minimal.
Preparing to unpack .../python2.7-minimal_2.7.12-1ubuntu0~16.04.9_amd64.deb ...
Unpacking python2.7-minimal (2.7.12-1ubuntu0~16.04.9) ...
Selecting previously unselected package python-minimal.
Preparing to unpack .../python-minimal_2.7.12-1~16.04_amd64.deb ...
Unpacking python-minimal (2.7.12-1~16.04) ...
```

Any errors will be displayed as stderr is not redirected to /dev/null

ci/build-docker-image.sh
ar13pit: Fixed in new commits

ci/install-package.sh
ar13pit: You mean fingerprint?

ci/install-package.sh
ar13pit: Updated this to be just a flag.

ci/install-package.sh
ar13pit: Fixed

.travis.yml
```diff
@@ -10,8 +10,14 @@ jobs:
     - &docker-generation
       stage: Docker generation
       language: generic
-      services: docker
       name: tue-env
+      before_install:
```
MatthijsBurgh: Have you tried the updated version from APT?
```yaml
addons:
  apt:
    packages:
      - docker-ce
```

ci/install-package.sh
MatthijsBurgh: Nope key signature, You can share it

.travis.yml
ar13pit: Yeah. It doesn't install the latest version. That's why I did this.

.travis.yml
ar13pit: Done.

.travis.yml
```diff
@@ -7,12 +7,20 @@ jobs:
...
       name: tue-env
-      script: ci/build-docker-image.sh tue-env
+      before_install:
+        - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+        - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+        - sudo apt-get update
+        - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
+        - sudo rm /etc/docker/daemon.json
+        - sudo systemctl restart docker
+      script: ci/build-docker-image.sh --image="tuerobotics/tue-env" --branch="$TRAVIS_BRANCH" --commit="$TRAVIS_COMMIT" --pull_request="$TRAVIS_PULL_REQUEST" --pull_request_branch="$TRAVIS_PULL_REQUEST_BRANCH" --user="$DOCKER_HUB_USERNAME" --password="$DOCKER_HUB_PASSWORD"
```
MatthijsBurgh: Add a space before starting script

ci/install-package.sh
ar13pit: You mean a string like this right?
```
12:f8:7e:78:61:b4:bf:e2:de:24:15:96:4e:d4:72:53
```

MatthijsBurgh: Yep

.travis.yml
ar13pit: Done.

ci/install-package.sh
ar13pit: Okay. Its also called fingerprint :)

.travis.yml
```diff
@@ -7,12 +7,21 @@ jobs:
...
       script: bash -c 'shopt -s globstar; shellcheck **/*.bash **/*.sh'
+
     - &docker-generation
       stage: Docker generation
       language: generic
-      services: docker
       name: tue-env
-      script: ci/build-docker-image.sh tue-env
+      before_install:
+        - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
MatthijsBurgh: Maybe better to put this in a script? So we could also use it in other travis file if needed for running the docker containers.

Dockerfile
ar13pit: Are you sure? Because if it does then all the things on the host file of travis will get added into the docker image. With this line we restrict it the hosts to what we want.

.travis.yml
ar13pit: Other travis files won't need this because only this is building the image while others are just pulling and executing a pre-built image.

ci/install-package.sh
ar13pit: Done.

Dockerfile
MatthijsBurgh: Are there may entries in the known hosts file? But people could overwrite the entire file to a limited set. But also we need to move it to the place where we use it. Which is not in tue-env

.travis.yml
MatthijsBurgh: Ok, but I still prefer a script.

ar13pit: Done.

Dockerfile
ar13pit: Yeah there a many entries in a known hosts file. Also as they are generated by ssh commands, there is no easy way of extracting a part from it.

I don't understand why you say we need to move this to somewhere else. This is only in the docker image not a part of tue-env in general. Doing this properly would mean integrating this in tue-env by parsing git url and checking if the ssh host exists in known host file, if not then adding it, which in my opinion is not in the scope of this PR.

ar13pit:
@MatthijsBurgh This PR is ready if don't have any more changes.

ar13pit:
But with this the CI of all repositories needs to be updated :crying_cat_face:

MatthijsBurgh:
Can we add default values so that the behavior stays the same without the CI changes? So we can change the CI configs gradually.

Dockerfile
MatthijsBurgh: This is what is in the Travis known_hosts file
https://travis-ci.com/github/tue-robotics/hero_bringup/jobs/297688538#L183

MatthijsBurgh: Azure doesn’t have a known hosts file by default

ar13pit:
> Can we add default values so that the behavior stays the same without the CI changes? So we can change the CI configs gradually.

This is added now.

ci/update-docker.bash
```diff
@@ -0,0 +1,24 @@
...
+    exit 1
+fi
+
+# Download latest stable docker version
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add - > /dev/null
+sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+sudo apt-get update -qq
+sudo apt-get -y -qq -o Dpkg::Options::="--force-confnew" install docker-ce > /dev/null
+
+# This is need on travis because of https://github.com/moby/moby/issues/39120 and https://github.com/travis-ci/travis-ci/issues/6418
```
MatthijsBurgh:
```suggestion
# This is needed on travis because of https://github.com/moby/moby/issues/39120 and https://github.com/travis-ci/travis-ci/issues/6418
```

ci/build-docker-image.sh
```diff
@@ -14,14 +14,44 @@ then
...
 
-IMAGE_NAME_SUBSTRING="$1"
-case $IMAGE_NAME_SUBSTRING in
+        -p=* | --password=* )
+            CI_DOCKER_PASSWORD="${i#*=}" ;;
+
+        -r=* | --registry=* )
+            CI_DOCKER_REGISTRY="${i#*=}" ;;
+
+        -s | --ssh )
```
MatthijsBurgh:
```suggestion
        --ssh )
```

ci/build-docker-image.sh
```diff
@@ -14,14 +14,44 @@ then
...
+            CI_DOCKER_SSH=true ;;
+
+        * )
+            echo -e "Error! Unknown input variable"
```
MatthijsBurgh:
```suggestion
            echo -e "Error! Unknown input variable '$i'"
```

Dockerfile
ar13pit: We could try mounting the file from the CI to docker container. That would resolve your issue with the explicit declaration of known hosts but I'll need to think how to simplify the build process. Mounting would work perfectly for `install-packages.sh` but in `build-docker-image.sh` that would be a problem. But we can skip it for now as we are not using ssh to build images.

ci/build-docker-image.sh
ar13pit: Seems like I missed it.

Dockerfile
MatthijsBurgh: Script to merge known_hosts file: https://blog.backslasher.net/merge-known-hosts.html

Dockerfile
```diff
@@ -49,7 +55,7 @@ RUN sed -e s/return//g -i ~/.bashrc && \
     export PULL_REQUEST=$PULL_REQUEST && \
     export COMMIT=$COMMIT && \
     # Run the standard installation script
-    source <(wget -q -O - https://raw.githubusercontent.com/tue-robotics/tue-env/"$BRANCH"/installer/bootstrap.bash) && \
+    source bootstrap.bash && \
```
MatthijsBurgh: Should we remove it after sourcing?

ar13pit: Doesn't really matter.

Dockerfile
MatthijsBurgh: Merging is added

ci/build-docker-image.sh
```diff
@@ -33,23 +63,37 @@ case $IMAGE_NAME_SUBSTRING in
...
+    cp ~/.ssh/known_hosts ./known_hosts
+else
+    touch ./known_hosts
+fi
 
-echo -e "\e[35m\e[1m Creating docker $IMAGE_NAME \e[0m"
+# Forward ssh-agent of the host
+if [ "$CI_DOCKER_SSH" == "true" ]
+then
+    DOCKER_SSH_ARGS="--ssh=default"
```
ar13pit: I was going to add exactly this to cleanup the docker build command further :)

ci/install-package.sh
```diff
@@ -78,12 +96,30 @@ echo -e "\e[35m\e[1m Trying to fetch docker image: $IMAGE_NAME:$BRANCH_TAG \e[0m
...
-    docker pull $IMAGE_NAME:$MASTER_TAG
-    BRANCH_TAG=master
+    docker pull "$IMAGE_NAME":"$MASTER_TAG"
+    BRANCH_TAG=$MASTER_TAG
+fi
+
+if [ -f ~/.ssh/known_hosts ]
+then
+    MERGE_KNOWN_HOSTS="true"
+    DOCKER_MOUNT_KNOWN_HOSTS_ARGS="--mount type=bind,source=$HOME/.ssh/known_hosts,target=/home/amigo/known_hosts_extra"
```
ar13pit: Do we really need `/home/amigo` here? Can it not be replaced with `~`

Dockerfile
```diff
@@ -28,19 +29,28 @@ SHELL ["/bin/bash", "-c"]
...
+    usermod -aG adm $USER && \
+    echo "%sudo ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/"$USER"
 
 # Setup the current user and its home directory
 USER "$USER"
 WORKDIR /home/"$USER"
 
+ADD installer/bootstrap.bash ./bootstrap.bash
+
+ADD ci/ssh-merge-known_hosts.py ./ssh-merge-known_hosts.py
```
ar13pit: As this file is not being used for docker build, there is no issue of adding it here. It automatically gets cloned in tue-env.

ci/install-package.sh
```diff
@@ -78,12 +96,30 @@ echo -e "\e[35m\e[1m Trying to fetch docker image: $IMAGE_NAME:$BRANCH_TAG \e[0m
...
 fi
 
 # Run the docker image along with setting new environment variables
-docker run --detach --interactive -e CI=true -e PACKAGE="$PACKAGE" -e BRANCH="$BRANCH" -e COMMIT="$COMMIT" -e PULL_REQUEST="$PULL_REQUEST" --name tue-env "$IMAGE_NAME:$BRANCH_TAG"
+# shellcheck disable=SC2086
+docker run --detach --interactive -e CI="true" -e PACKAGE="$PACKAGE" -e BRANCH="$BRANCH" -e COMMIT="$COMMIT" -e PULL_REQUEST="$PULL_REQUEST" --name tue-env $DOCKER_MOUNT_KNOWN_HOSTS_ARGS "$IMAGE_NAME:$BRANCH_TAG"
+
+if [ $MERGE_KNOWN_HOSTS == "true" ]
+then
+    docker exec tue-env bash -c "sudo chown 1000:1000 ~/known_hosts_extra && ./ssh-merge-known_hosts.py ~/.ssh/known_hosts ~/known_hosts_extra --output ~/.ssh/known_hosts"
```
ar13pit: Use the `ssh-merge-known_hosts.py` from tue-env and do not add it into the Dockerfile explicitly.

.dockerignore
```diff
@@ -2,3 +2,6 @@
...
+!ci/ssh-merge-known_hosts.py
```
ar13pit: Remove this line.

Dockerfile
MatthijsBurgh: True

ci/install-package.sh
ar13pit:
```suggestion
    DOCKER_MOUNT_KNOWN_HOSTS_ARGS="--mount type=bind,source=$HOME/.ssh/known_hosts,target=~/known_hosts_extra"
```

PR #477, opened 2020-03-29
More robust url switching

MatthijsBurgh:


setup/tue-functions.bash
```diff
@@ -230,29 +230,76 @@ complete -F __tue-git-clean-local _tue-git-clean-local
...
+    repo_address=${repo_address%.git}
+    echo -e "$domain_name\n$repo_address"
+}
+export -f _git_split_https # otherwise not available in sourced files
 
-    local web_address=${url#git@}
+function _git_split_ssh
+{
+    local url=$1
+    local web_address=${url#ssh://}
```
ar13pit, line 247: A url of this type will fail as domain_name will never be extracted.

setup/tue-functions.bash
```diff
@@ -230,29 +230,76 @@ complete -F __tue-git-clean-local _tue-git-clean-local
...
+    web_address=${web_address#git@}
     local domain_name=${web_address%%:*}
     local repo_address=${web_address#*:}
+    repo_address=${repo_address%.git}
+    echo -e "$domain_name\n$repo_address"
+}
+export -f _git_split_ssh # otherwise not available in sourced files
+
+function _git_split_url
```
ar13pit, line 256: Also, I don't really see the point of having 3 new functions when the changes you made could easily have been in `_git_https` and `_git_ssh`.

setup/tue-functions.bash
```diff
@@ -230,29 +230,76 @@ complete -F __tue-git-clean-local _tue-git-clean-local
...
+
+    echo "https://$domain_name/$repo_address.git"
 }
 export -f _git_https # otherwise not available in sourced files
 
 function _git_ssh
 {
     local url=$1
-    [[ "$url" == "git@"* ]] && echo "$url" && return 0
+    grep -q "^git@.*\.git$" <<< "$url" && echo "$url" && return 0
```
ar13pit, line 292: I'm a bit against grep when the input is known to be a single string.

```suggestion
    [[ $url =~ ^git@.*\.git$ ]] && echo "$url" && return 0
```
This is better IMO.

setup/tue-functions.bash
```diff
@@ -230,29 +230,76 @@ complete -F __tue-git-clean-local _tue-git-clean-local
...
+    fi
+    # shellcheck disable=SC2086
+    echo $output
+}
+export -f _git_split_url # otherwise not available in sourced files
+
+function _git_https
+{
+    local url=$1
+    grep -q "^https://.*\.git$" <<< "$url" && echo "$url" && return 0
```
ar13pit, line 275: Similar to comment below
```suggestion
    [[ $url =~ ^https://.*\.git$ ]] && echo "$url" && return 0
```

setup/tue-functions.bash
MatthijsBurgh:
```
_git_split_ssh ssh://git@github.com:tue-robotics/tue-env
result
github.com
tue-robotics/tue-env
```

setup/tue-functions.bash
MatthijsBurgh: `_git_split_url` is used on tue-install-impl in #478

PR #490, opened 2020-04-10
Pkg format 3, condition

MatthijsBurgh:
This PR continues on the work of #489, although it is not related. So review and merge it before reviewing this PR.

I still experience python2 to be faster over python3
```bash
time pythonX ~/.tue/installer/parse-package-xml.py ~/ros/$ROS_DISTRO/system/src/rosbridge_server/package.xml
```
python2: 28-34ms
python3: 49-73ms

STILL TO DO: Install python `regex` module for python version chosen to run the script.

P.S.:
Performance when using `evaluate condition from catkin_pkg.condition`
python2: 31-54ms
python3: 53-89ms
Although it doesn't support multiple consecutive `or` statements, mine implementation does.

Please take a look if performance could be improved.

Edit:
Condition can contain parenthesizes. `catkin_pkg` can already handle it, not yet released. I created a PR for a better solution, https://github.com/ros-infrastructure/catkin_pkg/pull/283.

Might be able to implement this also in current solution, otherwise decide to use `catkin_pkg`. Even with slower performance.


installer/parse-package-xml.py
```diff
@@ -0,0 +1,205 @@
...
+    }
+    assert parse_results[1] in operators.keys()
+    return operators[parse_results[1]](
+        _evaluate(parse_results[0], context),
+        _evaluate(parse_results[2], context))
+
+
+def main():
+    if len(sys.argv) != 2:
+        print("Usage: parse-ros-package-deps PACKAGE.XML")
```
ar13pit, line 164:
```suggestion
        print("Usage: parse-package-xml PACKAGE.XML")
```

MatthijsBurgh:
Replaced by #493, keep branch.

PR #498, opened 2020-04-25
Standardize docker image name

ar13pit:
- Replace `master` tag with `latest` for default branch as that is the docker nomenclature
- Use travis job name variable in .travis.yml
- Use colorized terminal in Docker

### DIFF
--- a/ci/build-docker-image.sh
+++ b/ci/build-docker-image.sh
@@ -63,7 +63,15 @@ case $image_substring in
 esac
 
 # Create tag based on branch name
-CI_DOCKER_IMAGE_NAME="$CI_DOCKER_IMAGE_NAME":$(echo "$CI_BRANCH" | tr '[:upper:]' '[:lower:]' | sed -e 's:/:_:g')
+default_branch=$(git remote show origin | grep HEAD | awk '{print $3;}')
+if [ "$default_branch" == "$CI_BRANCH" ]
+then
+    CI_DOCKER_IMAGE_TAG="latest"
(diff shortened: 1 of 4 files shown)


ci/build-docker-image.sh
```diff
@@ -63,7 +63,15 @@ case $image_substring in
 esac
 
 # Create tag based on branch name
-CI_DOCKER_IMAGE_NAME="$CI_DOCKER_IMAGE_NAME":$(echo "$CI_BRANCH" | tr '[:upper:]' '[:lower:]' | sed -e 's:/:_:g')
+default_branch=$(git remote show origin | grep HEAD | awk '{print $3;}')
```
MatthijsBurgh, line 66: Is the `;` need after the `$3`? As we don't have it in our `__tue-git-default-branch` function

MatthijsBurgh:
@ar13pit could we get the docker output colored?

ci/build-docker-image.sh
ar13pit: It is an optional syntax if there is only one command being executed. If multiple commands are executed then they must be separated by semi colons

ar13pit:
We can have colored output only during `docker run`. I haven't found a solution for during build.

MatthijsBurgh:
> We can have colored output only during `docker run`. I haven't found a solution for during build.

Lets do this in a separate PR.

ar13pit:
Too late. I pushed the commit :laughing:

PR #502, opened 2020-05-03
Refactor/tue status

MatthijsBurgh:
Still missing:
As a target would specify a branch, that should be ignored as well. We could either a file in the root of the git directory or store it somewhere else. But this file should also be removed, If the version is removed from the target.

I am not sure if this should already be implemented right now or could be done later.

### DIFF
--- a/setup/tue-functions.bash
+++ b/setup/tue-functions.bash
@@ -67,7 +67,10 @@ function tue-apt-select-mirror
 function _tue-git-get-default-branch
 {
     # Takes current dir in case $1 is empty
-    git -C "$1" symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'
+    local default_branch
+    default_branch=$(git -C "$1" symbolic-ref refs/remotes/origin/HEAD --short 2>/dev/null | sed 's@^origin/@@')
+    [ -z "$default_branch" ] && default_branch=$(git -C "$1" remote show origin 2>/dev/null | grep HEAD | awk '{print $3}')
+    echo "$default_branch"
 }
 
 function __tue-git-checkout-default-branch
--- a/setup/tue-functions.bash
+++ b/setup/tue-functions.bash
@@ -454,11 +454,16 @@ complete -F _tue-dev tue-dev
 function _robocup_branch_allowed
 {
     local branch=$1
-    if [ -f "$TUE_DIR"/user/config/robocup ] && [ "$branch" == "$(cat "$TUE_DIR"/user/config/robocup)" ]
-    then
-        return 0
-    fi
+    local robocup_branch
+    robocup_branch=$(_get_robocup_branch)
+    [ -n "$robocup_branch" ] && [ "$branch" == "$robocup_branch" ] && return 0
+    # else
     return 1
+}
+
+function _get_robocup_branch
+{
+    [ -f "$TUE_DIR"/user/config/robocup ] && cat "$TUE_DIR"/user/config/robocup
 }
 
 function _tue-repo-status
--- a/setup/tue-functions.bash
+++ b/setup/tue-functions.bash
@@ -497,13 +497,17 @@ function _tue-repo-status
             local current_branch
             current_branch=$(git -C "$pkg_dir" rev-parse --abbrev-ref HEAD)
 
-            local test_branches="master"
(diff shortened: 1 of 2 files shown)


setup/tue-functions.bash
```diff
@@ -64,16 +64,22 @@ function tue-apt-select-mirror
 #                                       GIT LOCAL HOUSEKEEPING
 # ----------------------------------------------------------------------------------------------------
 
-function __tue-git-default-branch
+function _tue-git-get-default-branch
+{
+    # Takes current dir in case $1 is empty
+    git -C "$1" symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'
```
ar13pit, line 70: Add a check: If this command fails, then use `git remote show origin | grep HEAD | awk '{print $3}'`

Creation of symbolic-ref is not very robust. I have seen it to be absent many times.

setup/tue-functions.bash
```diff
@@ -485,12 +496,28 @@ function _tue-repo-status
 
             local current_branch
             current_branch=$(git -C "$pkg_dir" rev-parse --abbrev-ref HEAD)
-            case $current_branch in
-                master|develop|indigo-devel|hydro-devel|jade-devel|kinetic-devel|toolchain-2.9) ;;
-                *) _robocup_branch_allowed "$current_branch" || echo -e "\033[1m$name\033[0m is on branch '$current_branch'";;
-            esac
-        fi
 
+            local test_branches="master"
```
ar13pit, line 500: Initialization to master is not needed.

setup/tue-functions.bash
```diff
@@ -485,12 +496,28 @@ function _tue-repo-status
...
+                    [ -z "$status" ] && return 0
+                    # else
+                    allowed="true"
+                    break
+                fi
+            done
+            [ "$allowed" != "true" ] && echo -e "\033[1m$name\033[0m is on branch '$current_branch'"
+        fi
         vctype=git
     elif [ -d "$pkg_dir"/.svn ]
```
ar13pit, line 522: Still want to keep this? None of our repos use svn

ar13pit:
Also I did not understand what you are trying to say in the missing part.

setup/tue-functions.bash
MatthijsBurgh: Based on being empty? Or by return code?

setup/tue-functions.bash
MatthijsBurgh: I don't see any reason to remove it right now, also not in scope of this PR.

setup/tue-functions.bash
ar13pit: Error code. When symbolic-ref fails, the error printed out is
```
fatal: ref refs/remotes/origin/HEAD is not a symbolic ref
```

setup/tue-functions.bash
ar13pit: Okay.

MatthijsBurgh:
> Also I did not understand what you are trying to say in the missing part.

For example, `kdl_parser`, the default branch of the repo is `melodic-devel`. But if you are on Noetic, the target specifies `noetic-devel`. But this will keep showing up in tue-status. This is not desired as tue-status should be empty in case nothing is modified.
So we should also allow branches defined in the target for that repo, also in case that branch isn't the default branch.

setup/tue-functions.bash
MatthijsBurgh: Fixed

ar13pit:
> > Also I did not understand what you are trying to say in the missing part.
> 
> For example, `kdl_parser`, the default branch of the repo is `melodic-devel`. But if you are on Noetic, the target specifies `noetic-devel`. But this will keep showing up in tue-status. This is not desired as tue-status should be empty in case nothing is modified.
> So we should also allow branches defined in the target for that repo, also in case that branch isn't the default branch.

I see. I think we should create a file in the workspace (.env directory) where repos with such a setting are listed out. tue-status checks this file and allows that branch.

ar13pit:
This looks good conceptually. But I haven't tested it. Have you tested it thoroughly? I can approve it then.

MatthijsBurgh:
I have tested it significantly, but it can wait till tomorrow.

PR #513, opened 2020-07-27
Add ...-now install target to nobleo get

Timple:


installer/parse-install-yaml.py
```diff
@@ -76,9 +76,8 @@ def main():
                 if "version" in install_item:
                     command += " {0}".format(install_item["version"])
 
-            elif install_type == "target" or install_type == "system" or install_type == "pip" or \
-                    install_type == "pip2" or install_type == "pip3" or install_type == "ppa" or \
-                    install_type == "snap" or install_type == "dpkg":
+            elif install_type in ["target", "system", "pip", "pip2", "pip3", "ppa", "snap", "dpkg",
+                                  "system-now", "pip-now", "pip2-now", "pip3-now", "ppa-now", "snap-now"]:
```
ar13pit, line 80: Haha! This is a nice hack :)

ar13pit:
> I think we need to update the readme as well.

Yeah. I'll do that and then merge this.

MatthijsBurgh:
I think the README can be more elegant. As we already have a table of all the functions. So Explaining them twice, is not that nice.

ar13pit:
> I think the README can be more elegant. As we already have a table of all the functions. So Explaining them twice, is not that nice.

The two things are complementary as I have not repeated everything from the table. So I think its fine, but if you can make it better then make a PR :)

README.md
```diff
@@ -119,6 +119,17 @@ Depending on Ubuntu distribution:
...
+#### (System / PIP / PIP2 / PIP3 / PPA / Snap)-now
+The default installation method for targets of type `system`, `pip(2/3)`, `ppa` and `snap` is to collect all such targets in a list and install them simultaneously at the end of the `tue-get install` procedure. To install such a dependency immediately for a specific target, use the target type as `X-now`:
+```yaml
+- type: [system/pip/pip2/pip3/ppa/snap]-now
+  name: <Name of the candidate>
+  
+- type: [target/system/pip/pip2/pip3/ppa/snap/dpkg]
+  name: <Name of the candidate>
+```
+or use `tue-install-X-now` functions in `install.bash` along with the procedure to installing the required target.
```
Timple, line 131: Actually this PR makes sure exactly this doesn't have to happen anymore :clown_face:

ar13pit: True. But it is still allowed to do so with `install.bash`. Maybe I should add that it is not recommended to do so with `install.bash` anymore?

Timple: Something like that would be nice.

PR #518, opened 2020-08-04
ci: Fix branch checkouts

ar13pit:
- Replace usage of environment variable `BRANCH` with positional arg `--branch` in `tue-install.bash` and `tue-install-impl.bash`
- In CI when running branch test, if the test branch is the default branch of the package then do not use the positional arg `--branch`
- Make default value empty for default branch in Dockerfile
- Set Docker image tags to source branch


ar13pit:
@MatthijsBurgh This fixes the issue of branch selection, however as a consequence of not using the environment variable `BRANCH`, tue-env-targets is always checked out to the master branch in the CI of tue-env. Not sure if that is okay.

However, that can be fixed by making a small change to the Dockerfile by adding `--branch` to the `tue-get install ros --test-depend` step.

MatthijsBurgh:
The targets need to be updated in install-packages to the correct branch is it happening?

Dockerfile
```diff
@@ -62,6 +62,8 @@ RUN --mount=type=ssh,uid=1000 sed -e s/return//g -i ~/.bashrc && \
     source ~/.bashrc && \
     # Set all git repositories to use HTTPS urls (Needed for local image builds)
     tue-env config ros-"$TUE_ROS_DISTRO" git-use-https && \
+    # Update branch of targets repository
```
MatthijsBurgh, line 65: Indent should be zero

Dockerfile
```diff
@@ -62,6 +62,8 @@ RUN --mount=type=ssh,uid=1000 sed -e s/return//g -i ~/.bashrc && \
...
+    tue-get update --branch="$BRANCH" && \
```
MatthijsBurgh, line 66: Why not add the Branch option to the `tue-get install ros`?

ci/install-package.sh
```diff
@@ -137,21 +137,20 @@ echo -e "\e[35m\e[1m ROS_DISTRO = ${ROS_DISTRO}\e[0m"
...
     # to their default branch.
     # This is needed before tue-get, so also new deps are installed.
     # After a tue-get run, we checkout forced, just to be sure.
 
-    # Install only the git repo of the package
-    echo -e "\e[35m\e[1m tue-get install ros-$PACKAGE --no-ros-deps\e[0m"
-    docker exec tue-env bash -c 'source ~/.bashrc; tue-get install ros-"$PACKAGE" --no-ros-deps'
-
-    # Fetch the merge branch
+    # Fetch the merged branch
```
MatthijsBurgh, line 153: Merge branch is correct. We are pulling the merge branch not the head branch of the PR.

ci/build-docker-image.sh
```diff
@@ -64,11 +64,14 @@ esac
 
 # Create tag based on branch name
 default_branch=$(git remote show origin | grep HEAD | awk '{print $3;}')
-if [ "$default_branch" == "$CI_BRANCH" ]
+CI_DOCKER_BRANCH=${CI_PULL_REQUEST_BRANCH:-$CI_BRANCH}
```
MatthijsBurgh, line 67: I think the pull request branch shouldn't be used at all. As the pull request test should test the merged state. So that is the feature branch merged into the target branch. So CI branch is the correct branch in that case.

Dockerfile
MatthijsBurgh, line 65: Indent should be zero

Dockerfile
MatthijsBurgh, line 66: Why not add the branch option to `tue-get install ros`? Also the branch option should be empty in case the branch is equal to the default branch.

installer/tue-install-impl.bash
```diff
@@ -1222,6 +1222,7 @@ shift
 
 # idiomatic parameter and option handling in sh
 targets=""
+BRANCH=
 while test $# -gt 0
```
MatthijsBurgh, line 1226: More consistent to use `BRANCH=""`?

ar13pit: I just kept it same as other parameters defined below the parameter handling loop. Can use your suggestion if you like for this line

Dockerfile
ar13pit: The branch option is equal to empty by default and the script build-docker-image passes an empty string in the default case. 

True, I could essentially just pass that arg to `tue-get install ros`, don't know what I was thinking last night :)

ci/install-package.sh
ar13pit: We are pulling the merged state which is the HEAD of the merged branch. The comment does not refer to PR branch anyway.

installer/tue-install-impl.bash
MatthijsBurgh: I just like some consistentcy. I don't have a preference for one or the other. But it conflicts with the targets line above it.

ar13pit: Haha your OCD is not allowing you to go past that line :rofl: 

I'll change it.

ci/build-docker-image.sh
```diff
@@ -64,11 +64,14 @@ esac
 
 # Create tag based on branch name
 default_branch=$(git remote show origin | grep HEAD | awk '{print $3;}')
```
MatthijsBurgh, line 66:
```suggestion
default_branch=$(git symbolic-ref refs/remotes/origin/HEAD --short 2>/dev/null | sed 's@^origin/@@')
[ -z "$default_branch" ] && default_branch=$(git remote show origin 2>/dev/null | grep HEAD | awk '{print $3}')
```

ci/build-docker-image.sh
ar13pit: Basically the CI_BRANCH should be checked out in the PR test as that if the branch the feature branch is being merged into. And if CI_BRANCH turns out to be default, then leave it empty. Pretty much what's being done in `install-package.sh` right?

MatthijsBurgh: Yes

PR #522, opened 2020-08-22
Add feature to only install doc or test deps

MatthijsBurgh:


installer/parse-package-xml.py
```diff
@@ -19,8 +19,11 @@ def main():
 
     dep_set = set()
 
-    dep_types = ['build_depend', 'buildtool_depend', 'build_export_depend', 'buildtool_export_depend', 'exec_depend',
-                 'depend', 'run_depend']
+    dep_types = []
+
+    if os.getenv('TUE_INSTALL_SKIP_ROS_DEPS', 'false') == 'false':
```
ar13pit, line 24: From what I see this variable either has a value `all` or `normal`, never `false`. Either this needs to change or the export value of the variable.

Preferably the export value because `all` and `normal` seem unusual for a `true` and `false` logic check. Unless there was a particular reason to choose them, which I'm not able to readily see here.

ar13pit (changes requested):

installer/parse-package-xml.py
MatthijsBurgh: I have taken a small shortcut here. Being bro defined is the same as false. An that is the same as not being all  or normal.

ar13pit (approved):

PR #524, opened 2020-11-01
Add target-now feature

MatthijsBurgh:
It installs the target and its dependencies directly, so these can be used during the execution of the install.bash script. Though ROS packages including dependencies, both system and git, are not converted to a now target, as ROS packages shouldn't be executed during install scripts.

I don't see the need for a command line option. As a target should only be converted to a now target in case that is needed by an other target.


ar13pit (approved):
LGTM. Try avoiding style diffs in feature PRs :).

PR #527, opened 2020-11-02
Correct error_handling of yaml install cmds

MatthijsBurgh:
If a target exists an error is not thrown because we want to check if the target with a ros prefix exists. But this causes a problem for not ros targets depending on other targets. Therefore the return code of cmds should be checked


ar13pit (approved):
Good catch :)

PR #528, opened 2020-11-02
Combine multiple now calls into one

MatthijsBurgh:


installer/parse-install-yaml.py
```diff
@@ -133,6 +147,12 @@ def main():
         command = command.replace(" ", "^")
         commands.append(command)
 
+    for install_type, pkg_list in now_cache.items():
```
ar13pit, line 150: This change doesn't preserve the order of the `-now` commands in the yaml file

ar13pit (changes requested):

installer/parse-install-yaml.py
MatthijsBurgh: That is true, but that shouldn't matter.

Because any item in the list should not depend on any other command in the list.

ar13pit: That is true

ar13pit (approved):

PR #535, opened 2020-11-28
Fix broken pipe in GitHub Actions

MatthijsBurgh:
Fixes https://github.com/tue-robotics/tue-robotics.github.io/runs/1466822946?check_suite_focus=true#step:3:63

### DIFF
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -983,9 +983,9 @@ function _tue-install-pip-now
     if [ -n "$pips_to_install" ]
     then
         echo -e "Going to run the following command:\n"
-        echo -e "yes | python${pv} -m pip install --user $pips_to_install\n"
+        echo -e "python${pv} -m pip install --user $pips_to_install <<< yes\n"
         # shellcheck disable=SC2048,SC2086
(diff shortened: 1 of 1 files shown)


ar13pit:
Why did this error not pop up before in Docker?

MatthijsBurgh:
Because I think this only happens in GH actions

ar13pit:
It is because of `-o pipefail` that has been set in shell setting. Before merging this, I would like to make the appropriate changes in the docker shell to see the same error.

MatthijsBurgh:
I think it happens without the pipefail too.

ar13pit:
It happens because of the options set on the bash shell, which are none in docker. In GH actions, the bash shell is running in strict mode

MatthijsBurgh:
I don't know what strict mode is, but the error happens with running just  `bash {0}`
Error: https://github.com/tue-robotics/tue-robotics.github.io/runs/1466840126?check_suite_focus=true#step:3:306, running  `bash {0}` https://github.com/tue-robotics/tue-robotics.github.io/runs/1466840126?check_suite_focus=true#step:3:9

ar13pit (approved):
LGTM

PR #539, opened 2020-12-10
Fix typo

MatthijsBurgh:


PetervDooren (commented):
Why ubuntu latest rather than the version we currently use? I get that this will be less maintenance in the future but I would prefer more control over the version we use.

MatthijsBurgh:
This only determines the ubuntu version for linting or the system that generates the docker. Not the ubuntu inside the docker container. That is still fully controlled

PetervDooren (approved):

PR #541, opened 2020-12-14
(actions) remove continue-on-error

MatthijsBurgh:
`continue-on-error` causes for the action to continue to run docker generation, while this shouldn't. This shouldn't have been there in the first place.

This fixes this incorrect behaviour: https://github.com/tue-robotics/tue-env/actions/runs/420599094


PetervDooren (approved):

PetervDooren:
The explanation was very much appreciated!

PR #560, opened 2021-03-23
Backport changes from Avular

ar13pit:


ar13pit:
@MatthijsBurgh I'm going to rebase Avular's master branch on top of tue-robotics' and make a new backport. This was done by a merge and I can see cleaning it up is a mess

installer/parse-install-yaml.py
```diff
@@ -88,6 +88,20 @@ def main():
                 else:
                     return show_error("Unknown ROS install type: '{0}'".format(source_type))
 
+            # Non ros targets that are packaged to be built with catkin
```
Timple, line 91: I get why `catkin` is a better description than `ros` for these kind of cmake packages.
But shouldn't the handling be exactly the same as a ROS target?

These cmake packages can actually have their own dependencies that have to be resolved as well. So the `package.xml` should be read and handled.

ar13pit: There is only one difference and that is system install. These cmake packages won't follow the rules of ROS packages. It is possible to use the same if block as that of "ros" but it would add another if block checking if the type is catkin, skip system install check. That becomes a lot of nested if else blocks

Timple: Than it would be nicer to grasp the common part in a function maybe?
This would add the dependency resolving for free to the catkin part.

MatthijsBurgh:
@ar13pit when are you planning to update this backport?

ar13pit:
@MatthijsBurgh Tonight :)

PR #561, opened 2021-03-26
Show environment usage

Timple:
and show `default` can also have a source

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,11 +137,11 @@ Taking the above into account, the following combinations for `install.yaml` are
        source:
          type: system
          name: <Package name>
-     default:
+     noetic:
        source:
          type: git
          url: <Repository URL>
-     noetic:
+     default:
        source: null
    ```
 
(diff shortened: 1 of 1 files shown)


MatthijsBurgh (dismissed):
Though I am not sure the comment about ROS2 is true. We didn't test that yet.

README.md
```diff
@@ -121,7 +140,7 @@ Taking the above into account, the following combinations for `install.yaml` are
        source:
          type: system
          name: <Package name>
-     indigo:
+     noetic:
```
Timple, line 143: Why did you flip it back from the default?

Now the example is not clear that `default` can also have something other than `pass`.

MatthijsBurgh: That was not clear from your changes.

MatthijsBurgh (approved):

README.md
Timple: I tried to catch that in the description:
> and show default can also have a source.

But failed I guess :smile:

MatthijsBurgh: Haha, 'read the manual' == 'read the commit message'

Timple: Well descriptive commit messages are important!

PR #562, opened 2021-03-26
Minor cosmetic changes

Timple:


installer/tue-install-impl.bash
```diff
@@ -588,7 +583,7 @@ Command: tue-install-cp $*"
             tue-install-debug "File $file and $cp_target are different, copying..."
             if "$root_required"
             then
-                tue-install-debug "Using elevated privileges (sudo)"
+                echo "Using elevated privileges (sudo) to copy ${file} to ${cp_target}"
```
MatthijsBurgh, line 586: Why not use `tue-install-info`?

installer/tue-install-impl.bash
```diff
@@ -1483,6 +1478,7 @@ then
     tue-install-snap-now "$TUE_INSTALL_SNAPS"
 fi
 
+
```
MatthijsBurgh, line 1481: Not needed IMO

installer/tue-install-impl.bash
```diff
@@ -472,21 +475,14 @@ function tue-install-hg
...
-    fi
-
-    tue-install-debug "Desired branch: $BRANCH"
-    if [ -n "$BRANCH" ] # Cannot be combined with version-if because this one might not exist
-    then
-        _try_branch_res=""
-        _try_branch_hg "$targetdir" "$BRANCH"
-        [ -n "$_try_branch_res" ] && res="${res:+${res} }$_try_branch_res"
-    fi
+    done
```
MatthijsBurgh, line 485: Also apply to `tue-install-git`

installer/tue-install-impl.bash
```diff
@@ -280,8 +281,10 @@ function _try_branch_git
         tue-install-error "Invalid _try_branch_git: needs two arguments (repo and branch)."
     fi
 
-    tue-install-debug "git -C $1 checkout $2"
-    _try_branch_res=$(git -C "$1" checkout "$2" 2>&1) # This is a "global" variable from tue-install-git
+    repo="$1"
+    branch="$2"
```
MatthijsBurgh, line 285: Use local variables?

installer/tue-install-impl.bash
```diff
@@ -280,8 +281,10 @@ function _try_branch_git
...
+    tue-install-debug "git -C $repo checkout $branch"
+    _try_branch_res=$(git -C "$repo" checkout "$branch" 2>&1)  # This is a "global" variable from tue-install-git
```
MatthijsBurgh, line 287:
```suggestion
    _try_branch_res=$(git -C "$repo" checkout "$branch" 2>&1)  # _try_branch_res is a "global" variable from tue-install-git
```

installer/tue-install-impl.bash
MatthijsBurgh: Apply this to `_try_branch_hg` too

installer/tue-install-impl.bash
MatthijsBurgh: Apply this to `_try_branch_hg` too

MatthijsBurgh (changes requested):

installer/tue-install-impl.bash
Timple: Glad you asked!

Because that is really intrusive. And you get a whole summary at the end as if something really cool just happened.

But the debug is too little, because `sudo` rights are asked and I have no clue why. That's why an echo just before _every_ sudo command is good practice I guess.

installer/tue-install-impl.bash
Timple: Didn't I? Whups

installer/tue-install-impl.bash
Timple: hg is not used in any target. Any reason not to remove the duplicate maintenance?

installer/tue-install-impl.bash
MatthijsBurgh: So we actually need a `tue-install-print` or something similar? As these echo's don't show in the logfile.

installer/tue-install-impl.bash
MatthijsBurgh: I am not sure about dropping HG support yet.

Timple: You know hg itself is deprecated?

MatthijsBurgh: I know bitbucket dropped support. But HG itself is still develop, correct?

Timple: Ah, I thought Atlassian 'owned' it.

Timple: But still, is it worth it? Do you really expect a ROS hg source in the future?

ar13pit: I explicitly removed hg, svn etc from Avular's fork as there is no point in maintaining something that's not being used anywhere

MatthijsBurgh: I will remove all HG features.

MatthijsBurgh:
I dropped support for SVN and HG. @Timple you only applied some changes to the HG branch functions. So that still needs to be applied to the Git functions.

installer/tue-install-impl.bash
```diff
@@ -280,8 +259,10 @@ function _try_branch_git
         tue-install-error "Invalid _try_branch_git: needs two arguments (repo and branch)."
     fi
 
-    tue-install-debug "git -C $1 checkout $2"
-    _try_branch_res=$(git -C "$1" checkout "$2" 2>&1) # This is a "global" variable from tue-install-git
+    repo="$1"
+    branch="$2"
```
MatthijsBurgh, line 263: Use local variables?

Remove quotes? I don't think these are needed. We don't use them in "arg parsing" of other functions.

MatthijsBurgh (changes requested):

Timple:
I'll pick this up before saturday!

installer/tue-install-impl.bash
Timple: Its in line with all the double line-endings surrounding the functions above :wink: 
It is a diff with our repo, not sure where it came from.

installer/tue-install-impl.bash
```diff
@@ -389,24 +392,19 @@ function tue-install-git
...
-        _try_branch_git "$targetdir" "$BRANCH"
-        [ -n "$_try_branch_res" ] && res="${res:+${res} }$_try_branch_res"
-    fi
+        if [ -n "$_try_branch_res" ]
+        then
+            res="${res:+${res} }$_try_branch_res"
+        else
+            rm "$version_cache_file" 2>/dev/null
+        fi
+    done
```
MatthijsBurgh, line 407: This is a change in behaviour to current setup. As in the current setup only the version specified in a target is stored in the cache file, not the version from `--branch`. I don't think we should do that. As I think that these versions should still show up in `tue-status.`

installer/tue-install-impl.bash
```diff
@@ -389,24 +392,19 @@ function tue-install-git
     tue-install-debug "Desired version: $version"
     local _try_branch_res # Will be used in _try_branch_git
     local version_cache_file="$TUE_ENV_DIR/.env/version_cache/$targetdir"
-    if [ -n "$version" ]
-    then
+    for branch in $version $BRANCH
```
MatthijsBurgh, line 395:
```suggestion
    for version in $version $BRANCH
```

MatthijsBurgh (changes requested):

installer/tue-install-impl.bash
Timple: Good point.

MatthijsBurgh:
Not needed after #563

Issue #351, opened 2019-01-18
Targets repo support


ar13pit:
- [ ] Add support for multiple targets repo for an environment
- [ ] Add support to associate the same targets repo with multiple environments
- [ ] Add support for different architectures of targets repo apart from git

PR #563, opened 2021-03-26
Allow for specifying multiple branches in order of specification

Timple:


MatthijsBurgh (changes requested):
Overlapping with #562?

Should `--branch` be removed? As it is removed from the docstring and implementation, but not from autocomplete.

Timple:
That would be a proper deprecation according to the tick-tock model :).

1. Support but don't encourage the old behavior.
2. Add deprecation warning **later**
3. Remove functionality

MatthijsBurgh:
Correction: not removed yet from implementation.

But according to your rules, you should remove it from the autocomplete IMO.

Timple:
Done, good to go

ar13pit:
This PR would probably break the CI scripts

Timple:
How?

ar13pit:
> How?

https://github.com/tue-robotics/tue-env/blob/master/ci/install-package.sh#L159

ar13pit:
My bad. It is still there in the implementation, so its not gone yet

Timple:
Yup, kept backwards compatibility.
@MatthijsBurgh you have a change-request still open which I believe is resolved.

ar13pit (dismissed):

ar13pit:
I think we should add the deprecated warning already for `--branch`

installer/tue-install-impl.bash
```diff
@@ -1366,6 +1366,9 @@ do
         --no-test-depend)
             export TUE_INSTALL_TEST_DEPEND="false"
             ;;
+        --branch*)
+            nobleo-install-info "Usage of --branch is deprecated, switch to --try-branch"
```
MatthijsBurgh, line 1370:
```suggestion
            tue-install-info "Usage of --branch is deprecated, switch to --try-branch"
```

installer/tue-install-impl.bash
```diff
@@ -1366,6 +1366,9 @@ do
...
+            ;;&
```
MatthijsBurgh, line 1371: What is this `&` doing here?

MatthijsBurgh (changes requested):

installer/tue-install-impl.bash
Timple: It continues the search in the case. So the branch option after this still works.

Timple:
All requested changes implemented I think.

MatthijsBurgh:
@Timple I prefer to merge #578 first. Then rebase this one and review once again.

ar13pit:
@MatthijsBurgh go ahead. Your PR has been merged

installer/tue-install-impl.bash
```diff
@@ -1229,6 +1232,14 @@ stamp=$(date_stamp)
 INSTALL_DETAILS_FILE=/tmp/tue-get-details-$stamp
 touch "$INSTALL_DETAILS_FILE"
 
+for branch in $BRANCH
+do
+    tue-install-debug "Parsed branch $branch"
+    # Switch tue-env immediately to this branch:
+    _try_branch "$TUE_DIR" "$branch"
+    [[ -n $_try_branch_res ]] && tue-install-info "[tue-env] $_try_branch_res"
```
MatthijsBurgh, line 1240: What `tue-env` doing here? Both the comment and the code

MatthijsBurgh (changes requested):

installer/tue-install-impl.bash
```diff
@@ -1229,6 +1232,14 @@ stamp=$(date_stamp)
...
+done
```
MatthijsBurgh, line 1241: Everything regarding `tue-env` should be handled in `tue-install.bash`

Right now I also get:
```
/home/amigo/.tue/installer/tue-install-impl.bash: line 1239: _try_branch: command not found
```

MatthijsBurgh (changes requested):

installer/tue-install-impl.bash
MatthijsBurgh: @Timple why do we check out `tue-env` to a different version? In the current setup `tue-env` is only pulled, never checked out. Not even in CI.

MatthijsBurgh (changes requested):

Timple:
Well, if you are working on a package that needs changes to `tue-env`, you can use the same branch name.
So CI succeeds while you are waiting for your PR to pass in `tue-env`.

Since targets are splitted here (we never got to that). The same logic should probably apply to `tue-env-targets` as well.

MatthijsBurgh:
Well that is already covered in the CI setup. We generate a docker image with the branch as tag and we use that in the CI, instead of master version of the docker image. So you only need to wait for the generation of that docker image.

MatthijsBurgh:
> Since targets are splitted here (we never got to that). The same logic should probably apply to `tue-env-targets` as well.

I really recommend splitting the targets.

Timple:
> We generate a docker image with the branch as tag

That leads to a large number of tags in the end. As they probably don't get cleaned up as the branch get's merged?

> I really recommend splitting the targets.

I don't see the benefit yet (except for it being nicely seperated). Until #351, that would be a real benefit.
That would also allow for 'personal' targets.

MatthijsBurgh:
> That leads to a large number of tags in the end. As they probably don't get cleaned up as the branch get's merged?

We do have a GH action which deletes the tags, https://github.com/tue-robotics/tue-env/blob/master/.github/workflows/delete_docker_tags.yml

Yes #351, would be really nice.

MatthijsBurgh:
I am not sure it is easy to implement, but best would be to check the branches in reverse order. And if stop after the first checkout which succeeds. This would reduce the number of checkouts.

MatthijsBurgh (approved):

MatthijsBurgh:
I assumed the last one in the argument list has the highest priority. So when that one is available, there is no need to check the other ones before.
In the old situation, you would start by the branch with lowest priority and checkout that one. And you kept checking out branches with a higher priority. This way you would end with the available branch with the highest priority.

So by going in reverse order of the argument list, so start with the highest priority and yo stop when you found a branch available. This way you still end-up with the branch with highest priority, but with less tries.

Timple:
Ahaa! That does make sense indeed!

MatthijsBurgh:
@Timple could you please review my changes.

Timple:
I'm not allowed to approve my own PR :upside_down_face: 
But yes, agreed.

installer/tue-install-impl.bash
```diff
@@ -444,13 +449,17 @@ function tue-install-git
...
+        tue-install-debug "Parsed branch '${branch}'"
         _try_branch_res=""
-        _try_branch_git "$targetdir" "$BRANCH"
-        [ -n "$_try_branch_res" ] && res="${res:+${res}\n}$_try_branch_res"
-    fi
+        _try_branch_git "${targetdir}" "${branch}"
+        _try_branch_error_code=$?
+        [ -n "${_try_branch_res}" ] && res="${res:+${res}\n}${_try_branch_res}"
+        [ "${_try_branch_error_code}" -eq 0 ] && break
+    done
```
ar13pit, line 462: I think we should add a note/caution somewhere in the README/code that this behavior has changed. Consider this: We have the branch list to be `foo bar baz`. If one the repos has both `foo` and `bar` as available branches in it, then the repo is checked out with `foo` and not `bar`. So having multiple branches specified might result in unintended behavior if not used correctly.

ar13pit (changes requested):

installer/tue-install-impl.bash
MatthijsBurgh: No because of the reverse order inside the script, it will be checked out to `bar`. But it is good to mention the priority order of the arguments in the docs.

ar13pit (approved):

ar13pit:
Maybe squash and merge

MatthijsBurgh:
I think I might squash some together, but not all.

PR #564, opened 2021-03-26
Future proofing

Timple:


setup/tue-functions.bash
```diff
@@ -328,7 +328,7 @@ function tue-make
     if [ -n "$TUE_ROS_DISTRO" ] && [ -d "$TUE_SYSTEM_DIR" ]
     then
         local build_tool=""
-        if [ -f "$TUE_SYSTEM_DIR"/devel/.built_by ]
+        if [ -f "$TUE_SYSTEM_DIR"/build/.built_by ]
```
Timple, line 331: Not all build systems (colcon) have a `devel` dir. But all have a `build` dir.

installer/tue-install-impl.bash
```diff
@@ -101,7 +101,12 @@ function tue-install-target
     tue-install-debug "tue-install-target $*"
 
     local target=$1
-    local now=$2
+
+    if [[ "$ROS_PYTHON_VERSION" == 3 && "$target" == python-* ]]
+    then
+        tue-install-error "Trying to install a python2 target (${target}) while the ROS default is python3. Fix your dependencies"
+        return 3
+    fi
```
MatthijsBurgh, line 109: How about `python-*-pip` targets. Not all will be migrated to `python3-*-pip` in rosdep, which we follow.

setup/tue-functions.bash
```diff
@@ -838,8 +838,9 @@ function tue-get
     shift
 
     #Create btrfs snapshot if possible and usefull:
-    if [[ "$cmd" =~ ^(install|update|remove)$ ]] && { df --print-type / | grep -q btrfs; }
+    if [[ -z "BTRFS_SNAPSHOT" && "$cmd" =~ ^(install|update|remove)$ ]] && { df --print-type / | grep -q btrfs; }
```
MatthijsBurgh, line 841: Please include in README.

MatthijsBurgh (changes requested):
Please fix linting.

installer/tue-install-impl.bash
Timple: Simply open a PR there, they get accepted easily :slightly_smiling_face: 

But sure, you can remove the check if it doesn't suit the tooling.

MatthijsBurgh: I did, https://github.com/ros/rosdistro/pull/28264. So they are not accepting python3 versions of existing `python-*-pip` targets. And are not accepting new `python-*-pip` targets either, only `python3-*-pip` (https://github.com/ros/rosdistro/pull/28672). So you get a big mess of `python-*-pip` and `python3-*-pip`.

Timple: Hah, this way the deprecation will never go away because people can't use the new format.

Alrighty then... This should be gone.

MatthijsBurgh (dismissed):
Please wait with merge for approval of @ar13pit

ar13pit (approved):
LGTM

setup/tue-functions.bash
```diff
@@ -838,8 +838,9 @@ function tue-get
     shift
 
     #Create btrfs snapshot if possible and usefull:
-    if [[ "$cmd" =~ ^(install|update|remove)$ ]] && { df --print-type / | grep -q btrfs; }
+    if [[ -z "$BTRFS_SNAPSHOT" && "$cmd" =~ ^(install|update|remove)$ ]] && { df --print-type / | grep -q btrfs; }
```
ar13pit, line 841: Shouldn't this be -n and not -z?

ar13pit (changes requested):

setup/tue-functions.bash
ar13pit: @Timple how does this btrfs snapshot work because I have never seen it work? Also the '/' in a standard ubuntu installation is ext4 so the second check in the if block always fails

Timple: It only works on an BTRFS filesystem (can be chosen when you install ubuntu). It can make instant snapshots which don't take any space until you modify files again.

Timple: `{ df --print-type / | grep -q btrfs; }` this will be false on your PC then.

MatthijsBurgh: @Timple how about the `-z` or `-n`?

Timple: @ar13pit is right, should be -n.

Probably didn't catch is because of the other bug (found by linter) which covered this bug :)

Timple:
```suggestion
    if [[ -n "$BTRFS_SNAPSHOT" && "$cmd" =~ ^(install|update|remove)$ ]] && { df --print-type / | grep -q btrfs; }
```

MatthijsBurgh (approved):

PR #573, opened 2021-06-08
Do not check versions if nothing needs installing

Timple:
If all targets end up in the `git_pips_to_install` variable, the `pips_to_check` variable will be empty. And crash the version-check script.

### DIFF
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -1021,36 +1021,39 @@ function _tue-install-pip-now
         fi
     done
 
-    read -r -a pips_to_check <<< "$pips_to_check"
-    local installed_versions
-    if [ ${#pips_to_check[@]} -gt 0 ]
+    if [ -n "$pips_to_check" ]
     then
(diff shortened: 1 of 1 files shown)


installer/tue-install-impl.bash
```diff
@@ -1021,36 +1021,39 @@ function _tue-install-pip-now
         fi
     done
 
-    read -r -a pips_to_check <<< "$pips_to_check"
-    local installed_versions
-    if [ ${#pips_to_check[@]} -gt 0 ]
+    if [ -n "pips_to_check" ]
```
Timple, line 1024: The diff is a mess. Basically I put all version-check logic behind this `if`.

MatthijsBurgh:
Is this not already catched by https://github.com/tue-robotics/tue-env/pull/573/files#diff-d61f6ec58e4025f247404b8280fba24427c392ae4aa7f60dcd34ce34687d33e4L1026?

Timple:
Whups, we're out of date I guess...

PR #575, opened 2021-07-02
Add rosdep support for missing targets

Timple:
To test: simply remove `ros-std_msgs` from your targets. That should normally break stuff I guess :sunglasses:

### DIFF
--- a/installer/parse-install-yaml.py
+++ b/installer/parse-install-yaml.py
@@ -107,6 +107,7 @@ def main():
                 "pip3-now",
                 "ppa-now",
                 "snap-now",
+                "rosdep",
             ]:
                 if now and "now" not in install_type:
                     install_type += "-now"
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -984,7 +984,7 @@ function tue-install-snap-now
 
 function tue-install-rosdep
 {
-    tue-install-debug "tue-install-rosdep $@"
+    tue-install-debug "tue-install-rosdep $*"
 
     if [ -z "$1" ]
     then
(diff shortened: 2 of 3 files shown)


Timple:
This cleaned up 80% of our targets and reduces a lot of maintenance effort which is done in the rosdep repository anyway.

Timple:
Ow sorry, accidentally commited the `unset-default` option.

Let me know if you need this splitted.

MatthijsBurgh:
I also gave this a try in the past, https://github.com/tue-robotics/tue-env/pull/385. That version does support pip. I think that solution is a bit more elegant. Please let me know what you think.

Timple:
pip support should be trivial to add here as well.

Your implementation only resolves using rosdep and than tries to install those again using own functions.

This implemenation treats rosdep as any other package manager (apt, pip, snap). So for code consistency I like this better.

MatthijsBurgh:
I don't think that is better way to see rosdep. I see rosdep as a resolving tool. Not as a package manager. What I also like about my solution is that it errors much earlier.

Timple:
That's semantics I guess ;-)

Don't apt/pip/snap also do a lookup-and-install? Just like rosdep?

I'm sure you can also lookup what apt would do and then install with dpkg. But that's just cumbersome.

MatthijsBurgh:
But then you should not parse the rosdep output to forward it to apt. Then you should let rosdep do the call to apt. Then I think it fits in the package manager list

For my the considerations in this decision are:
- What is the fastest?
- I prefer to keep the number of calls to apt-get limited.
- Error as early as possible in a way that we can print the target that causes the error.
- (Keep `tue-get dep` working)

Timple:
> But then you should not parse the rosdep output to forward it to apt. 

Ah, I thought we didn't do that, my bad.

Ahwel. This works. Feel free to modify/improve/reject.

PR #581, opened 2021-07-26
Make clear that branch is checked out

MatthijsBurgh:
Fix issue of https://github.com/tue-robotics/tue-env/runs/3159129462?check_suite_focus=true#step:4:131

### DIFF
--- a/installer/tue-install-impl.bash
+++ b/installer/tue-install-impl.bash
@@ -291,8 +291,8 @@ function _try_branch_git
         tue-install-error "Invalid _try_branch_git: needs two arguments (repo and branch)."
     fi
 
-    tue-install-debug "git -C $1 checkout $2"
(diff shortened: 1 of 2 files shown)


MatthijsBurgh:
Later in the CI it keeps waiting to re-open stdin. Same as when not using `< /dev/null`, see https://github.com/tue-robotics/tue-env/runs/3159179565?check_suite_focus=true

The error msg has been there for a longer time. (Last master run https://github.com/tue-robotics/tue-env/runs/3017548690?check_suite_focus=true#step:4:702). So If we fix that issue, we can eliminate the `< /dev/null`.

ar13pit:
Do you know which apt package is requiring the dpkg-preconfigure execution?

MatthijsBurgh:
> Do you know which apt package is requiring the dpkg-preconfigure execution?

Nope

MatthijsBurgh:
This was caused by the circular dependency. See #582 for the real solution.

PR #585, opened 2021-08-04
Fix CI for azure/travis

MatthijsBurgh:
Get rid of unnecessary spacing in logging.

### DIFF
--- a/ci/set-branch.sh
+++ b/ci/set-branch.sh
@@ -4,9 +4,6 @@ BRANCH=""
 for i in "$@"
 do
     case $i in
-        -r=* | --repo=* )
-            REPO="${i#*=}"
-        ;;
         -b=* | --branch=* )
             BRANCH="${i#*=}"
         ;;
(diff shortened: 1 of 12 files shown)


ci/set-branch.sh
```diff
@@ -0,0 +1,42 @@
...
+    esac
+    shift
+done
+
+echo -e "\e[35m\e[1mREPO   = ${REPO}\e[0m"
+echo -e "\e[35m\e[1mBRANCH = ${BRANCH}\e[0m"
+
+echo -e "\e[35m\e[1mGetting default branch by GitHub API of: ${REPO}\e[0m"
+
+DEFAULT_BRANCH=$(curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/"${REPO}" 2>/dev/null | jq ".default_branch" -r)
```
ar13pit, line 29: Why not use the git repo itself to get the default branch? That would free us from using the Github API.

ar13pit (changes requested):

ci/set-branch.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [[ "$BRANCH" == "$DEFAULT_BRANCH" ]]
+then
+    echo -e "\e[35m\e[1mOn the default branch, clearing BRANCH variable\e[0m"
+    BRANCH=""
+else
+    echo -e "\e[35m\e[1mOn a feature branch, keeping the BRANCH variable: ${BRANCH}\e[0m"
+fi
+
+export BRANCh
```
ar13pit, line 34: Nice typo @MatthijsBurgh :laughing: 
```suggestion
export BRANCH
```

ar13pit (changes requested):

ci/set-branch.sh
MatthijsBurgh: Keeping you sharp ;)

ar13pit:
My only question now is why not use the same logic in GitHub Actions?

MatthijsBurgh:
> y not use the same logic i

This would require cloning of the repo in single package repo's, which we don't do now. This would require changing the current configs

ar13pit (approved):

PR #604, opened 2022-02-15
Add BSD-3 License

### LINKED ISSUE #603: No license defined
There's no license definition in this repository. For that reason, it's not possible to re-use any of the code in other repositories.

We encountered this in practice with @MatthijsBurgh when he suggested to use [ci/commit-range/action.yml](https://github.com/tue-robotics/tue-env/blob/master/ci/commit-range/action.yml) in the [Darker](https://github.com/akaihola/darker) project (see akaihola/darker#260). I can't copy that GitHub action definition to the Darker repository since there's no license.

MatthijsBurgh:
We have BSD-2 in other repos, but I like the 3rd clause.

Fixes #603


PetervDooren (approved):

ar13pit:
I like the addition

PR #607, opened 2022-03-05
Support pip options

MatthijsBurgh:


installer/tue-install-impl.bash
```diff
@@ -810,15 +810,18 @@ function _tue-install-pip
     local pv=$1
     shift
     tue-install-debug "tue-install-pip${pv} $*"
+    local name="$*"
+    name="${name// /^}"
```
ar13pit, line 814: name has been substituted here and below??

ar13pit (changes requested):

installer/tue-install-impl.bash
MatthijsBurgh: Forgot that

ar13pit (approved):

PR #612, opened 2022-05-13
Cache ccache

MatthijsBurgh:
For GH actions users, it doesn't require any changes.

Without cache[4m:05]: https://github.com/tue-robotics/ed_python/actions/runs/2319312444/attempts/4
With cache[1m:45]: https://github.com/tue-robotics/ed_python/actions/runs/2319312444/attempts/8

In theory the cache should not affect the outcome of the build process. As it should not be used when files are changed. But we should keep an eye on it.


ar13pit (approved):
Squash and merge with version update

PR #616, opened 2022-08-15
Support gems

MatthijsBurgh:


installer/tue-install-impl.bash
```diff
@@ -1048,6 +1048,55 @@ function tue-install-snap-now
...
+            tue-install-debug "gem pkg: $pkg is not yet installed"
+        else
+            tue-install-debug "gems pkg: $pkg is already installed"
+        fi
+    done
+
+    if [ -n "$gems_to_install" ]
+    then
+        # shellcheck disable=SC2086
+        tue-install-pipe sudo gem install $gems_to_install || tue-install-error "An error occurred while installing gem packages." #<<< yes
```
Timple, line 1094: Does `gem` always require `sudo`?

Timple (approved):

installer/tue-install-impl.bash
MatthijsBurgh: That was how it was used in the gitsu target. And running `gem install gitsu` does indeed return a permission error.

PR #621, opened 2022-09-29
Generate target_setup.bash file in python

MatthijsBurgh:


setup/generate_setup_file.py
```diff
@@ -0,0 +1,74 @@
...
+
+import os
+import sys
+import traceback
+
+
+class SetupGenerator:
+    def __init__(self, tue_env_dir=None):
+        if tue_env_dir is None:
+            tue_env_dir = os.environ["TUE_ENV_DIR"]
```
ar13pit, line 13: raise exception if this is still None

setup/generate_setup_file.py
```diff
@@ -0,0 +1,74 @@
...
+        self._tue_env_dir = tue_env_dir
+        self._tue_dependencies_dir = os.path.join(self._tue_env_dir, ".env", "dependencies")
+        self._tue_env_targets_dir = os.environ["TUE_ENV_TARGETS_DIR"]
```
ar13pit, line 16: raise exception if this is None

setup/generate_setup_file.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self._visited_targets = set()
+
+    def generate_setup_file(self):
```
ar13pit, line 20:
```suggestion
    def generate_setup_file(self) -> None:
```

setup/generate_setup_file.py
```diff
@@ -0,0 +1,74 @@
...
+            lines.extend(self._generate_setup_file_rec(dep))
+
+        target_setup_file = os.path.join(self._tue_env_targets_dir, target, "setup")
+        if os.path.isfile(target_setup_file):
+            lines.append(f"source {target_setup_file}\n")
+
+        return lines
+
+
+def main():
```
ar13pit, line 62:
```suggestion
def main() -> int:
```

ar13pit (changes requested):

setup/generate_setup_file.py
MatthijsBurgh: I think this will raise a `KeyError` when not existing. When the user sets it to an empty string. Than it is useless, but the script still works.

setup/generate_setup_file.py
MatthijsBurgh: Again, this will raise a `KeyError`, when not defined.

ar13pit (approved):

PR #629, opened 2022-11-29
fix(git) remove quotes to resolve tilde

MatthijsBurgh:


ar13pit:
This is not correct. We should make sure that instead of using "~" we use "$HOME" in the input because removing the quotes will not protect from path splitting.

PR #638, opened 2023-01-03
WIP: migrate tue-get to python

MatthijsBurgh:

PR #639, opened 2023-01-07
Feature/versioning test

PetervDooren:
Add python script to check whether the versions of existing packages match those specified in a packagexml


PetervDooren:
development on this will continue after #638 is largely done

PR #672, opened 2023-05-23
feat: AGMAH backport

ar13pit:


installer/parse_install_yaml.py
```diff
@@ -164,6 +163,7 @@ def get_distro_item(item: Mapping, key: str, release_version: str, release_type:
                 return {"system_packages": system_packages, "commands": commands}
 
             elif install_type == "ros":
+                ros_release = environ["TUE_ROS_DISTRO"]
```
MatthijsBurgh, line 166: Now this is done multiple times, instead of once.

MatthijsBurgh (changes requested):

installer/parse_install_yaml.py
ar13pit: Having it global has a different implication. If TUE_ROS_DISTRO is not set, then our whole tool breaks. This variable is only required while installing a ros system package

MatthijsBurgh: But I think we require TUE_ROS_DISTRO to be set anyway in multiple spots.

ar13pit: We don't. Until ros packages are installed, it is not required

MatthijsBurgh: But we do always set it in the `boostrap.bash`

.github/workflows/main.yml
```diff
@@ -118,12 +118,14 @@ jobs:
           echo "tag=${IMAGE_TAG}" >> $GITHUB_OUTPUT
       - name: Generate artifacts
         id: artifacts
+        if: github.event_name != 'pull_request'
```
MatthijsBurgh, line 121: Should we do this on "worklflow_dispatch"?

The same for all added conditions

MatthijsBurgh (changes requested):
I need explanation on 89808c47b90da33d9febd06e166a0f4df347bb39

.github/workflows/main.yml
```diff
@@ -32,18 +36,33 @@ jobs:
...
+      - name: Login to Docker registry
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install QEMU
+        run: |
+          sudo apt update -qq
+          sudo apt install -qq -y qemu qemu-user-static
```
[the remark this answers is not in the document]
ar13pit: This is on the standard Ubuntu installation without custom dependencies so if the step fails, it is very clear because of which package. In this case, having `-qq` makes sense.

installer/parse_install_yaml.py
ar13pit: I know and that becomes a problem when running multiple distros on the same machine. My fix to this to have only tue-env setup on the base computer, create different tue-env workspaces which have docker compose scripts specific to that workspace and then run docker containers from them where TUE_ROS_DISTRO is setup through the bootstrap script.

MatthijsBurgh (changes requested):
We need more logging in the new CI code

MatthijsBurgh:
> > We need more logging in the new CI code
> 
> Can you be a bit more specific about where? There's not much which has been added. It is just a job that is combining multiple manifest files of different architectures :)

I think we should minimally log everything we write to either GITHUB_ENV or GITHUB_OUTPUT

ar13pit:
This PR actually removes most of the untagged containers from being created to begin with. All the `unknown/unknown` ones to be precise.

MatthijsBurgh:
> This PR actually removes of the random untagged containers from being created to begin with. All the `unknown/unknown` ones to be precise.

That is great improvement. But the old job still pushes the BRANCH-amd64 tag. And the manifest job pushes the BRANCH tag. So we should streamline this worflow.

MatthijsBurgh:
@ar13pit can you work on the logging? So that at least everything that is written to variables etc are also printed, human-readable please.

.github/workflows/main.yml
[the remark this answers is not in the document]
MatthijsBurgh: This is only needed when generating ARM64

MatthijsBurgh (approved):

PR #742, opened 2024-06-10
(CI) Improve SSH stuff

MatthijsBurgh:
Fixes the SSH issues in CI, see https://github.com/tue-robotics/hero_bringup/actions/runs/9447499194/job/26019423899


dockerfiles/tue-env.Dockerfile
```diff
@@ -66,6 +66,7 @@ WORKDIR /home/"$USER"
 RUN mkdir -p -m 0700 ~/.ssh
 ADD ./known_hosts ./.ssh/known_hosts
 RUN sudo chown $USER_ID:$USER_ID ~/.ssh/known_hosts && sudo chmod 644 ~/.ssh/known_hosts
+RUN cp ~/.ssh/known_hosts ~/.ssh/known_hosts.bak
```
ar13pit, line 69: Why this? Generation of this docker image happens at the very end of the build script and this cp and mv won't be required anywhere

ar13pit (changes requested):

dockerfiles/tue-env.Dockerfile
MatthijsBurgh: I added this, because when you build the docker, the installation can add additional hosts to the known host file. I am not sure, but it could be this was causing issues, when using the docker in CI again.

MatthijsBurgh: I tested it with the lines removed, but that way the `known_hosts` is extend during building the docker. This is undesired IMO. So the docker is as clean as possible.

PR #751, opened 2024-10-23
Add pre-commit config, incl. fixes

MatthijsBurgh:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,90 @@
...
+      #   ]
+      - id: no-commit-to-branch
+      - id: pretty-format-json
+      # - id: requirements-txt-fixer
+      # - id: sort-simple-yaml
+      #   files: []
+      - id: trailing-whitespace
+        args: [--markdown-linebreak-ext=md]
+
+  - repo: https://github.com/psf/black
```
Timple, line 62: I would use `ruff` here. As it has the same formatting but also does `isort` and `flake8` at once.

MatthijsBurgh: Thanks, I work with ruff too. Switching to ruff just requires a bit more work. Which I will do in a separate PR. I already started on it a while back.

